PR #20, opened 2018-10-22
doc: add license and CoC

MylesBorins:
these were missing. Same boilerplate as other repos


MylesBorins:
Hey all... as this is a housekeeping bit and having the license in particular is important to allow people to contribute I would like to land this EOD, if there are no objections

PR #22, opened 2018-10-22
[WIP] Proposal for Foundation Membership

jasnell:


MylesBorins:
I do not personally see it being scalable to have multiple projects with board representation. It is my current belief that the need for board representation could be solved better by having groups such as a Cross Project Council be given broader decision making authority and budget.

While I think have a multi-tiered system for projects makes sense, I don't like the idea of it mirroring the board structure of Platinum, Gold, Silver as I feel it creates the wrong incentive structures. I would much rather see something like Sandbox, Incubator, Top-Level. There should be a clear way for projects to get higher level representation... although this should not by default be the board.

It is also of my opinion that Node.js should not be grandfathered into any special lifetime status. I recognize that this is going to be controversial, but seeing the sentiment towards projects grandfathered into the JSF has left people feeling it seems like doing such would start the foundation off on the wrong foot. Node.js should definitely be a model for decided what a Top-Level project looks like, but I do not think it would be healthy for a new foundation to create a special status for Node.js. If the Node.js is large and well governed it should always be able to remain a top-level project. If the project's governance corrodes or it stops being used as much it should not maintain special status.

I will not block or -1 this PR as it is explicitly a proposal and tied to an individual, but wanted to raise my objections in here for historical purposes

jasnell:
> I do not personally see it being scalable to have multiple projects with board representation. It is my current belief that the need for board representation could be solved better by having groups such as a Cross Project Council be given broader decision making authority and budget.

What is needed here to resolve this is precisely what @rubys pointed out: Node.js top level committee members clearly believe that Node.js needs a seat at the Foundation board but it's not yet clear *why*. At the face to face in Vancouver I also spoke with @groundwater who also felt that Electron, should it decide to move into such a foundation, would also likely want a seat on the Board. Again, the key question is not, "Is that scalable", but "Why would such a seat be necessary in the first place?".

What I think you'll find once we start digging into the reasons, is that it'll come down to three very specific points:

1. Brand management per Project
2. Budget Allocation per Project
3. Resource Allocation per Project

Addressing the concerns around these will go a long way towards resolving any issues around the perceived need for a seat at the board table.

> While I think have a multi-tiered system for projects makes sense, I don't like the idea of it mirroring the board structure of Platinum, Gold, Silver as I feel it creates the wrong incentive structures. I would much rather see something like Sandbox, Incubator, Top-Level. There should be a clear way for projects to get higher level representation... although this should not by default be the board.

Not sure I follow with regards to "wrong incentive structures". Can you clarify the concern there.

I don't particular care what the tiers are called with regards to projects, the overall effect is the same and the details of the proposal don't change.

> It is also of my opinion that Node.js should not be grandfathered into any special lifetime status... If the Node.js is large and well governed it should always be able to remain a top-level project. If the project's governance corrodes or it stops being used as much it should not maintain special status.

That being said then, the criteria for what would qualify (or disqualify) Node.js (and others) as "Top-Level Projects" needs to be clarified as one of the specific deliverables of this committee. There is a non-trivial number of Node.js top level committee members who feel strongly about the fate of Node.js not being in the arbitrarily defined hands of individuals who are not directly contributing to the project, so a clear set of goal posts must be defined.

groundwater:
This full scope of this is a little beyond me, but I'll fill in _some_ of my reasoning:

1. the foundation wants more platinum members
2. there are large organizations (not just GitHub) bought into Electron
3. having a temporary seat (e.g. ~2yr) is an excellent way of demonstrating value for them to buy in at a platinum level
4. at the expiry of those 2yr, you _ideally_ have new members

This is not my entire position, but I thought it was worth sharing the above perspective.

rubys:
As James suggested, I believe that board membership is solving the wrong problem.  As the Node.js board is currently the place where Brand Management, Budget Allocation, and Resource Allocation takes place, there naturally is a desire for a seat at that table.

This isn't scaleable.  In software engineering terms, a refactoring is in order.

I suggest that the ASF provides one such model.  One that has hundreds of projects and dozens of committees.  With a mix of paid staff and volunteers.

Taking budget as one example.  The ASF board approves (or ratifies) a budget, but doesn't prepare one.  The President (who isn't a necessarily a board member, and in fact, currently isn't) collects up and prepares an annual budget request by working with the chair of each committee (or project).  Once the budget is approved, the owner of a line item manages that money during the course of the year, and the board gets monthly reports tracking spending to plan.

The split isn't precisely as James described.  Instead of each project having a Trademarks budget, there is a Brand Management committee that manages that budget, and that is a President's committee not a board committee.  Membership of that committee is managed pretty much like membership of the Modules committee is done in Node.js - i.e., without board involvement.  So people that are interested in that topic come together.

Similar things for Travel Assistance and Conferences and other committees.

I believe that such an approach would work well here, and would solve a number of problems.

MylesBorins:
@jasnell are you still planning to pursue this or can it be closed?

PR #23, opened 2018-10-22
Draft Staging Process

jorydotcom:
Context: On our call today, I proposed borrowing the idea of a staging process from TC-39 to our bootstrap program and policy ideas work.

This PR adds a simple staging-process.md doc for discussion, which we can iterate on and flesh out with more examples. 

Applying the proposed process, this PR would be in Stage 0. Assuming it is in the right place and there are enough +1s, we should flesh this out into a stage 1 proposal between now and the next meeting.

### DIFF
--- a/STAGING_PROCESS.md
+++ b/STAGING_PROCESS.md
@@ -1,0 +1,17 @@
+# Staging Process Proposal
+
+Anyone may submit an idea for a policy or program following the staging process outlined below. To submit an idea, create a proposal document in markdown format in the /proposals directory, update [proposals/README.md](./README.md), submit a pull-request, and apply the appropriate staging label. If the proposal has multiple artifacts please consider making a folder for the proposal.
+
+The proposal will be evaluated at bootstrap committee meetings and stage changes will occur at those times.
+
+## Stage 0 - Initial Proposal
+Proposals in this stage are summarized in a markdown file. They have a short name and an identified 'champion' - at least one person who will be pushing the proposal through to its conclusion. They have a rough description of the program idea or policy, what resources would be needed to execute it, what committee or role is responsible for overseeing it, ideas for metrics, execution, and so forth. They have clear language about which group or groups the program/policy will serve. The purpose of this stage is to capture the spirit of an idea the community may want to explore further, and to begin thinking about implementation ideas or process questions that need to be answered.
+
+## Stage 1 - Draft Proposal 
+Proposals in this stage are written up in the style of the formal policy or program language. They include examples of the program idea or policy in action, with clear steps and guidelines. They include the steps that will need to be taken to initially implement the idea or policy. The purpose of this stage is to refine the policy, make sure all stakeholders (including the board) have had a chance to review and request changes, and determine whether further iteration is needed. 
+
+## Stage 2 - Accepted Proposal 
+Proposals in this stage are nearly complete. They may be in a pilot or trial-run mode for a limited audience, or pending legal review. They may be awaiting budget approval. The purpose of this stage is to complete any operational checks, such as acquiring budget, access, or Board approval if needed, before launching the program.
+
+## Stage 3 - Complete
+The proposal is now an adopted policy or program of the Foundation, and is owned by the committee or role identified. It is expected that the policy or program will be re-evaluated on some regular basis and deprecated if need be.
--- a/proposals/TEMPLATE.md
+++ b/proposals/TEMPLATE.md
@@ -26,3 +26,7 @@ Please include potential metrics that can be used to measure success of the prop
 ## Why this proposal is important
 
 Why should we do this?
+
+## What is neccessary to complete this proposal
+
+Define "done".
--- a/proposals/TEMPLATE.md
+++ b/proposals/TEMPLATE.md
@@ -1,0 +1,28 @@
+# My New Proposal
+>  Stage 0
+
+Below is simply a suggested template for drafting a proposal, it will not fit all potential proposals.
+
+## Champion
+
+Who is the the stakeholder of this initiative?
+
+## Description
+
+A rough description of the program idea or policy
+
+## Required Resources
+
+What resources would be needed to execute it?
+
+## Who would be responsible?
+
+Which committee would be overseeing or responsible for executing this proposal
+
+## How would success be measured?
+
+Please include potential metrics that can be used to measure success of the proposal.
+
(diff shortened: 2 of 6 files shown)


mhdawson:
I'd agree that a bit more info as to where in the repo docs in each of the stages should go would be helpful

staging-process.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Stage 2 - Draft 
+Proposals in this stage are written up in the style of a policy. They include examples of the program idea or policy in action, with clear steps and guidelines. They include the steps that will need to be taken to initially implement the idea or policy. The purpose of this stage is to refine the policy, make sure all stakeholders (including the board) have had a chance to review and request changes, and determine whether further iteration is needed. 
+
+## Stage 3 - Candidate 
+Proposals in this stage are near complete. They may be in a pilot or trial-run mode for a limited audience, or pending legal review. They may be awaiting budget approval. The purpose of this stage is to complete any operational checks, such as acquiring budget, access, or Board approval if needed, before launching the program.
+
+## Stage 4 - Complete
+The proposal is now an adopted policy or program of the Foundation, and is owned by the committee or role identified. It is expected that the policy or program will be re-evaluated on some regular basis and deprecated if need be.
```
jasnell: may I suggest a simpler process.

* Stage 0 - Initial proposal
* Stage 1 - The committee is reviewing the proposal
* Stage 2 - The proposal is being incorporated into the feedback provided by the committee

If there are competing Stage 1 proposals with no clear consensus, then both are combined into a single Stage 2 proposal that discusses both as options.

staging-process.md
```diff
@@ -0,0 +1,19 @@
+Anyone may submit an idea for a policy or program following the staging process outlined below. To submit an idea, create a pull-request and apply the appropriate staging label.
```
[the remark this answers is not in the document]
jorydotcom: okiedok!

staging-process.md
jorydotcom: That makes sense to me; does there need to be some process to figure out when it can go forward or what to do in a stalemate situation?

Related, should there be a timeline for proposals? Say something hasn't moved from Stage 0 in a year, it gets moved to an archive or folder or something?

jorydotcom:
@mhdawson done!

I added a title & some language about where proposals could live. Also I consolidated the stages a bit - one thing I'm not sure about is that since we don't know what the committee process/budget process will look like I feel like we might get into a spot where some project committee could say, "this proposal is good to go" but we still need some kind of sign-off from the c3 committee for example before it's done done done. So I left that last stage in there for now just to capture that concern & see where everyone is at on that.

staging-process.md
```diff
@@ -0,0 +1,17 @@
+# Staging Process Proposal
+
+Anyone may submit an idea for a policy or program following the staging process outlined below. To submit an idea, create a proposal document in markdown format in the /proposals directory, submit a pull-request, and apply the appropriate staging label. 
+
```
mhdawson: Should we ask that the stage be in the title or have the markdown files live in subdirectories for example: /proposals/stage-0

staging-process.md
jasnell: The stalemate situation is covered the last sentence of my comment there. Specifically:

> If there are competing Stage 1 proposals with no clear consensus, then both are combined into a single Stage 2 proposal that discusses both as options.

MylesBorins:
I'm a big +1 of this. Small structural nit

I think this might be better if we have a proposals folder and include a tiny template. I've implemented an example in https://github.com/MylesBorins/bootstrap/commit/5ee5bfb9b7e837d4d93d60cdb6027ec6d1691220

Check it out at https://github.com/MylesBorins/bootstrap/tree/my-bootstrap/proposals

LMK if you would like a PR. This PR was built on top of https://github.com/nodejs/bootstrap/pull/7/, but I think it would make more sense to rebase against master and be more self contained

MylesBorins:
Hey all. After discussing with @jorydotcom I've gone ahead and rebased this PR against master... updated the staging process to reflect the discussion in the last meeting, and refactored my GOVERNANCE proposal to reflect the new documented process.

PTAL and comment away. It would be very nice to land this in the next meeting. We can always iterate and refactor as things move along

jorydotcom:
this looks good to me, @MylesBorins! I like adding a quick look at the proposals in the readme and providing a template, that will be really helpful to folks I think. Thanks for carrying this forward last week ❤️

STAGING_PROCESS.md
```diff
@@ -0,0 +1,17 @@
...
+
+The proposal will be evaluated at bootstrap committee meetings and stage changes will occur at those times.
+
+## Stage 0 - Initial Proposal
+Proposals in this stage are summarized in a markdown file. They have a short name and an identified a 'champion' - at least one person who will be pushing the proposal through to its conclusion. They have a rough description of the program idea or policy, what resources would be needed to execute it, what committee or role is responsible for overseeing it, ideas for metrics, execution, and so forth. They have clear language about which group or groups the program/policy will serve. The purpose of this stage is to capture the spirit of an idea the community may want to explore further, and to begin thinking about implementation ideas or process questions that need to be answered.
```
thefourtheye: an identified 'champion'

proposals/TEMPLATE.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Who would be responsible?
+
+Which committee would be overseeing or responsible for executing this proposal
+
+## How would success be measured?
+
+Please include potential metrics that can be used to measure success of the proposal.
+
+## Why this proposal is important
```
thefourtheye: Nit: Question mark at the end.

proposals/mylesborins-GOVERNANCE/README.md
```diff
@@ -0,0 +1,27 @@
+# Initial Governance Model
+>  Stage 0
+
+## Champion
+
+Myles Borins (@Myles Borins)
+
+## Description
+
+This document is a basic outline of the governance model for the bootstrap team. It outline how membership is defined and how we will collaborate.
```
thefourtheye: It outlines...

proposals/TEMPLATE.md
MylesBorins: As it is currently written this is a statement not a questions
`Why this proposal is important` vs `Why is this proposal important`

MylesBorins:
landed in 8574222...500ea22

PR #27, opened 2018-10-26
Andy

rubys:
From Andrew Updegrove:

> Gentlemen, here are some thoughts/notes on how it might make the most sense to move forward on merging the bylaws from a process efficiency point of view.
>
> Attached is a comparison of the bylaws of the two orgs on which I've color coded the different categories of changes to suggest how I would address them, as follows:
>
> - Gray – administrative changes
> - Yellow – corporate member concerns
> - Green – tracks statute/boilerplate
>
> While all changes are marked for transparency, my thought is that this draft would be useful by indicating what changes are likely to be of interest to the community and what may not. In the attached draft, the unshaded areas would be the ones to focus on. My assumption is that working off of this draft would save everybody a lot of time, which still allowing anyone to make their own judgment and, if they wish, still raise an issue under one of the shaded portions.
>
> If you think it would be helpful, I'd be happy to go through the unshaded diffs and, where it seems like people might not be likely to have a strong opinion one way or another, make a suggestion which way the language should come out (this would include differences that are more likely to be random than significant). I'd also be happy to make a recommendation on which approach I think is better for other items based on the over 100 organizations I've represented that have all started working from the same template bylaws (as the template has evolved over the last 30 years). This might help make the process a bit shorter and less painful for everybody.

[Compare & Categorize Node.js vs JS Foundation Bylaws - GU 10-25-18.docx](https://github.com/nodejs/bootstrap/files/2517301/Compare.Categorize.Node.js.vs.JS.Foundation.Bylaws.-.GU.10-25-18.docx)

[Compare & Categorize Node.js vs JS Foundation Bylaws - GU 10-25-18.pdf](https://github.com/nodejs/bootstrap/files/2517302/Compare.Categorize.Node.js.vs.JS.Foundation.Bylaws.-.GU.10-25-18.pdf)


MylesBorins:
@rubys does this make sense to land as either a proposal or an artifact rather than in the root directory?

rubys:
If there were a directory for proposals, it would indeed make sense for this file to go in there.

MylesBorins:
@rubys fleshing out that process in https://github.com/nodejs/bootstrap/pull/23

rubys:
@MylesBorins I'll just note that there doesn't appear to be any disagreement w.r.t Stage 0 and Stage 1, so it doesn't make sense to me to block proposals from advancing to either of these levels until the definitions of higher level stages are resolved.

MylesBorins:
@rubys I rebased https://github.com/nodejs/bootstrap/pull/32 against https://github.com/nodejs/bootstrap/pull/23 which seems very close to be landing. Would you perhaps want to attempt to use the process documented in that PR?

MylesBorins:
Can this be closed

MylesBorins:
One more ping to @rubys 

If there isn't a response by the 28th I'm going to close (it can always be reopened)

PR #36, opened 2018-11-12
doc: move governance to stage 2

MylesBorins:
As there is no need for board, budgetary, or legal approval
I believe this proposal can immediately go to stage 3

If there are no objections I'll modify this PR to do so.


sarahnovotny:
LGTM

PR #48, opened 2018-11-23
doc: initial version of CPC Charter

mhdawson:


mhdawson:
I think main areas for discussion are:

* Composition of CPC (Section 4).  There are 3 different models when looking at Node.js TSC, JS Foundation TAC and CNCF TOC.  They are quite different in terms of how much involvement the Foundation members have in selecting the members.  There are also differences with respect to terms. both JS and CNCF have a limit on membership before needs to be re-confirmed, versus the Node.js TSC where this is not the case.

* Mandate.  I think what is there based on pulling in content from the 3 is reasonable but this is an area
  we should discuss expand in terms of some of the "new" things proposed as part of the merged
  foundation. For example "Commons.js"

I could put together a suggestion for what I think makes in terms of CPC membership, but I believe some discussion on that point would be good and then I can update to reflect that discussion. 
 I've left some sections with \> \> \> \> \> \> \> \> \> \> where I have the content form 2 or more of Node, JS and CNCF so people can see the different approaches.

mhdawson:
Adding bootstrap agenda so that we can have a discussion in an upcoming meeting.  @MylesBorins would be great if we could get it on this weeks agenda as I'm going to miss the follow 2 weeks....

proposals/mdawson-cpc-charter/CPC-Charter.md
```diff
@@ -0,0 +1,305 @@
...
+
+   * ensuring collaboration is the driving principle within a Project, between
+     _$FOUNDATION_ Projects, and between _$FOUNDATION_ Projects and the broader
+     community.
+   * defining and maintaining neutral consensus for the technical vision for the _$FOUNDATION_
+   * approving new projects within the scope for _$FOUNDATION_ set by the Governing Board,
+     as outlined in [XXXX], and creating conceptual architecture for how the proejcts
+     fit together.
+   * aligning projects, removing or archiving projects, as outlined in [XXXX].
+   * defining common practices to be implemented across _$FOUNDATION_ projects, if any.
```
mcollina: I would remove this entirely.

proposals/mdawson-cpc-charter/CPC-Charter.md
```diff
@@ -0,0 +1,305 @@
...
+   * The CPC is also responsible for mediating technical conflicts between
+     _$FOUNDATION_ Projects when attempts to resolve those conflicts within the Project
+    and were unsuccessful.
```
mcollina: Can you define what a technical conflict might be? What would be the goal of that mediation?

proposals/mdawson-cpc-charter/CPC-Charter.md
```diff
@@ -0,0 +1,305 @@
...
+     as outlined in [XXXX], and creating conceptual architecture for how the proejcts
+     fit together.
+   * aligning projects, removing or archiving projects, as outlined in [XXXX].
+   * defining common practices to be implemented across _$FOUNDATION_ projects, if any.
+   * The CPC is also responsible for mediating technical conflicts between
+     _$FOUNDATION_ Projects when attempts to resolve those conflicts within the Project
+    and were unsuccessful. 
+   * The CPC serves as the _$FOUNDATION_'s primary technical
+     liaison body with external open source projects, consortiums and groups and is
+     also responsible for technical participation in standards bodies
```
mcollina: This has a slight implication that, in order to participate as $FOUNDATION to standard bodies, you have to be a member of CPC. I think that would be a bit elitist. IMHO somebody might be interested in being part of the CPC but not participate in standards and vice versa.

I think this responsibility should be delegated to a subcommitee/or somehow worded differently to enable such delegation.

proposals/mdawson-cpc-charter/CPC-Charter.md
```diff
@@ -0,0 +1,305 @@
...
+   * Members of the CPC, or committees consisting of subsets of the CPC
+     membership will be expected to meet on a regular basis to discuss topics
+     such as new project acceptance into the mentorship program, project graduation
+     from the mentorship program, Project issues and conflicts, opportunities
+     for collaboration between Projects, opportunities for the Foundation in the
+     greater JavaScript community, etc.
+
```
mcollina: I think this is missing something related to _funds_, the travel budget, and things like that. Or are those under the C3C?

One of the goal of the CPC is to follow closely Board/$FOUNDATION initiatives and communicate them to the projects.

proposals/mdawson-cpc-charter/CPC-Charter.md
mhdawson: I would read it more that the CPC has the responsibility to make sure that the right things happen, not that you have to be on the CPC itself.

mcollina: Then we should probably align the test, this seems to be a direct responsibility.

proposals/mdawson-cpc-charter/CPC-Charter.md
mhdawson: I think it might be useful to either leave it in, or move it to the non-goals so that we know we have agreement one way or the other.

proposals/mdawson-cpc-charter/CPC-Charter.md
mhdawson: I took this directly from the responsibilities of the JSF TAC (along with the bad wording :).  A good example does not come to mind, but then I've not been working in a Foundation with multiple projects. @jorydotcom do you have any examples you can add?

Will fix the wording though.

proposals/mdawson-cpc-charter/CPC-Charter.md
mhdawson: Updated the text, let me know if that sounds better.

proposals/mdawson-cpc-charter/CPC-Charter.md
mhdawson: Added those 2 above.

mhdawson: Added those a bit further up in the list.

MylesBorins:
Is this perhaps far enough along to move to stage 1 instead of stage 0?

Do we want to try and land something on Monday?

mhdawson:
As long as there are no glaring contradictions with the discussions that have been held on the roles,  I think moving to stage 1 makes sense to me.  Would it make sense to land as stage 0 and then file PR to move to stage 1?  I think I'd need at least one approval to do that.

MylesBorins:
@mhdawson I think we can land as stage 1. maybe present it to the group in today's meeting and then land later today if there are no objection?

PR #64, opened 2018-12-18
proposal: SHARED_INTERFACES documentation effort

littledan:
["JavaScript Shared interfaces"](https://github.com/littledan/js-shared-interfaces/) is a documentation effort to explain how cross-environment ("isomorphic") JavaScript code can be written. My hope is that this can help convergence between different platforms over time.


mhdawson:
I think this is a good example of the kind of activities that would be great to sheppard under the merged foundation.

MylesBorins:
landed in 61b10543d4

PR #68, opened 2019-01-07
doc: 2019-01-07 notes

MylesBorins:
Refs: https://github.com/nodejs/bootstrap/issues/67

### DIFF
--- a/meetings/2019-01-07.md
+++ b/meetings/2019-01-07.md
@@ -1,0 +1,115 @@
+# Joint Foundation Bootstrap Team Meeting 2019-01-07
+## Links
+
+* **Recording**: https://www.youtube.com/watch?v=FCYlOxYWSlg
+* **GitHub Issue**: https://github.com/nodejs/bootstrap/issues/67
+
+## Present
+
+- Myles Borins (@MylesBorins)
+- Tierney Cyren (@bnb)
+- Todd Moore (@tmmoore_1)
+- Michael Dawson (@mhdawson)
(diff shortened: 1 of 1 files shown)


meetings/2019-01-07.md
```diff
@@ -0,0 +1,115 @@
...
+    - What does C3 look like? [#38](https://github.com/nodejs/bootstrap/issues/38)
+    - What does the CPC look like? [#37](https://github.com/nodejs/bootstrap/issues/37)
+    - Top Level Council(s) Shape and Focus [#52](https://github.com/nodejs/bootstrap/issues/52)
+    - https://github.com/nodejs/bootstrap/projects/1
+    - [CPC Charter Proposal](https://github.com/nodejs/bootstrap/blob/master/proposals/mdawson-cpc-charter/README.md)
+
+  - Myles: Recap work done to differentiate CPC and C3 responsibilities in project board. Not
+    clear we need a hard seperation between CPC and C3. This is important for us to figure out 
+    sooner rather than later because it will affect allocation of board seats and by-laws.
+  - c
```
mhdawson: - Myles asked if there were any objections to moving forward with one group which would be the CPC.  
   No objections from those present, but will ask again in next meeting as not everybody can make each
   of the time slots.
- Myles: Does Michael's CPC charter cover board representation? If we just have the CPC then we need to
  factor that in.
- Michael: Not, yet. Will take action to create PR to add that in. (PR is now here: 
   https://github.com/nodejs/bootstrap/pull/70)

PR #70, opened 2019-01-08
Boardmembers2

mhdawson:
Note that is is cumulative and includes changes from an outstanding PR. Please just look at the last commit: https://github.com/nodejs/bootstrap/pull/70/commits/a401a7b3b8a6991b9c8d6d848edd952cda673764

### DIFF
--- a/proposals/mdawson-cpc-charter/CPC-Charter.md
+++ b/proposals/mdawson-cpc-charter/CPC-Charter.md
@@ -53,7 +53,7 @@ The CPC is in inclusive group with 2 types of members:
 as outlined below. In any case where it is not specified that a role
 or responsibility applies to a Voting member, then a Regular member is
 implied. Anyone who has been a member of one of the projects under the 
-foundation for at least 3 months may request to
+Foundation for at least 3 months may request to
 become a regular member by opening a PR to add themselves to the list of
 regular members. Regular members remain for as long as they are active
 within the work of the CPC. Regular members who have not been active
--- a/proposals/mdawson-cpc-charter/CPC-Charter.md
+++ b/proposals/mdawson-cpc-charter/CPC-Charter.md
@@ -52,25 +52,25 @@ The CPC is in inclusive group with 2 types of members:
 **Regular members** Regular members have the roles and responsibilities
 as outlined below. In any case where it is not specified that a role
 or responsibility applies to a Voting member, then a Regular member is
-implied. Any member of one of the projects under the foundation may request to
+implied. Anyone who has been a member of one of the projects under the 
+foundation for at least 3 months may request to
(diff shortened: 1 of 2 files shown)


MylesBorins:
Had a discussion with @hackygolucky about the proposed individual membership program

If we are considering still having a program we may want to augment this proposal to have only a single seat delegated by the CPC, as the second seat would be managed by the individual membership

mhdawson:
@MylesBorins I did have that thought as well. I left it as is through because the individual membership program has a threshold of 2000 members which I thought we might not hit in the first year.  I was thinking that once that threshold was hit then the CPC could update the processes to assign one of the 2 board members to the Individual membership program.

MylesBorins:
@mhdawson the challenge would be what happens in the interim. Does one of the directors have to resign to make space for the individual membership director? what happens if it activates in month 13... do we wait 11 months to bring someone in?

mhdawson:
I would say that it waits until the next year.

MylesBorins:
@mhdawson that seems very odd to me tbqh

brianwarner:
If it helps provide context, the way this works in the Node bylaws for the other directors is that the board composition doesn't change until the terms have been completed.  For example, the Gold and Silver seats are pegged to the number of Platinum directors.  If the number of Platinum directors falls below the threshold for the current number of Gold directors mid-year, the existing Gold directors serve out their term and then the next election reflects the recalculated number.

mhdawson:
I think changing on the yearly basis is what makes sense to me and is consistent with @brianwarner described

hackygolucky:
@mhdawson @brianwarner that doesn't seem to parallel the challenge we're seeing here because there's -always- some representation for those member tiers, yes? For Silver or Gold as an example in the corporate class, it just adds more seats when the threshold is met. The silver members never go without representation, just less. What we're talking about here is making a class of members go unrepresented for a given amount of time by a director of their choosing and accountable to them until a director with other focuses times out on the seat.

brianwarner:
In the current bylaws, it varies between Silver and Gold.  There is always at least one Gold director, so long as there is at least one organization in the Gold membership class.  The Silver director seats only activate when there are 10 or more Silver members.

MylesBorins:
@brianwarner but the situation you are describing is when it goes below not above. So there wouldn't be a case where the 10th silver member joins and 11 months pass without representation

mhdawson:
I think waiting until the next year is ok.  If its ok not to have a director at 1999, then when we hit 2001 it is ok to wait until the next election period for there to be one. In addition, there would to be some lead time anyway to put the process for elections in place, find candidates etc. so there is likely to be a couple of month delay at a minimum anyway.

MylesBorins:
landed in e4f893ecec

To be updated in another PR as discussed in today's meeting

PR #71, opened 2019-01-10
doc: re-organize proposals

MylesBorins:
This re-organizes the proposal doing the following

* removes champion name from proposal name
* organizes based on stages
* cleanup of folder names for consistency


MylesBorins:
I've requested review from all individuals whose proposals would be affected by this

PR #72, opened 2019-01-10
proposal: Community Board Representation Stage 1

MylesBorins:
TL;DR

* 1 CPC board seat
* 1 floating board seat
  - allocated by CPC annually
  - Requires 2 Platinum members to activate
* 1 Individual Membership board seat
  - Requires 3 Platinum members and 2k Individual Members to activate

I believe the README.md has everything necessary to reach
consensus on this, thus going for stage one. It will need to be
implemented in the CPC Charter and Foundation Bylaws

### DIFF
--- a/proposals/stage-0/stage-1/community-board-representation/README.md
+++ b/proposals/stage-0/stage-1/community-board-representation/README.md
@@ -1,0 +1,64 @@
+# Community Board Representation
+>  Stage 1
+
+## Champion
+
+[Myles Borins](https://github.com/MylesBorins)
+
+## Description
+
+### What exists today
+
+Currently the Node.js foundation has up to 3 community board seats:
+
+* 1 board seat for TSC
+* 1 board seat for Community Committee
+* 1 board seat for Individual Membership Program
+
+There are some limitations on these seats as noted in [Section 4.3](https://github.com/nodejs/board/blob/master/by-laws.md#section-43-nomination-election-and-term-of-office-of-directors),
+specifically that:
+
+4.3 d) There must be 3 platinum directors for the individual membership theo elect a board member
+
+4.3 e) There must be 2 platinum directors for the TSC or CommComm to elect a board member
+
+### Proposal for $FOUNDATION
+
+There would be up to 3 community board seats:
+
+* 1 board seat for CPC
+* 1 floating board seat to be delegated by CPC
+  - delegated on an annual basis and approved by Board
+* 1 board seat for the individual membership program
+
+Similar to the current bylaws there would be some limitations that would activate seats:
+
+* CPC would always have a board seat independent of platinum membership
+* There must be 2 platinum directors to activate the floating board seat
+* There must be 3 platinum directors and 2k individual members to activate the individual board member seat
+
+These limitations allow for representation to scale with the growth of the foundation while also ensuring that there
+will always be community representation.
+
+## Who would be responsible?
+
+This would be owned by both the board and the CPC
+
+## Why this proposal is important
+
+Community representation is extremely important to the success of the foundation ensuring the board has insight
+into the community and vice versa.
+
+## Unresolved Question
+
+* Will there be an individual membership program?
+* Is the community ok with the proposed allocation
+* Is the community ok with 1 of the seats only activating with 2k members
+* Is the board ok with the proposed allocation and limitations
+
+## What is necessary to complete this proposal
+
+* Community consensus
+* Board Sign off
+* Implemented in CPC Charter
+* Implemented in $FOUNDATION bylaws
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -11,6 +11,7 @@ Anyone may submit an idea for a policy or program following the [staging process
 * [Expectations](./stage-1/EXPECTATIONS)
 * [Project Progression](./stage-1/PROJECT_PROGRESSION)
 * [CPC Charter](./stage-1/CPC_CHARTER)
+* [Community Board representation](./stage-1/community-board-represenation)
 
 ## Stage 2
 


proposals/stage-0/stage-1/community-board-representation/README.md
```diff
@@ -0,0 +1,64 @@
...
+4.3 d) There must be 3 platinum directors for the individual membership theo elect a board member
+
+4.3 e) There must be 2 platinum directors for the TSC or CommComm to elect a board member
+
+### Proposal for $FOUNDATION
+
+There would be up to 3 community board seats:
+
+* 1 board seat for CPC
+* 1 floating board seat to be delegated by CPC
```
mhdawson: The challenge I see here is how we address the discussion around representation for the Node.js project.  With 2 it is easier to have a split which addresses the issue as outlined in: https://github.com/nodejs/bootstrap/pull/70. Any thoughts of how we address the issue if there is only one seat.  We could just have a vote from all members and since Node.js has more members then it would be able to select a representative.  Downside is that other projects may not feel fully represented. On the other hand it does not seem fair to set up an approach for other smaller projects with less total members could select the representative over that the Node.js community might want.

The other advantage is that if for some reason the individual membership program does not workout or grow, the CPC would have the flexibility to chose some other process for the second seat.

MylesBorins: The vote would be only by voting members of CPC afaik... meaning node.js would be only 2 votes of arguably between 6 - 12 votes. This would be the case for anything that the CPC would vote on.

Re: floating seats vs non floating. If the membership program doesn't work we can always adjust bylaws.. this is something we did re: commcomm. I think that having the seat be activated by number of members will be a great motivation for the projects in the foundation to help the program be successful

mhdawson: I don't think that is the way it works for CNCF, its a vote among all of the members.  If it is the only CPC members then we are back to the discussion about Board representation for the Node.js project (either confirming that there is/is not an expectation on that front as has been expressed and if there is how to respond to that).

mhdawson:
@mcollina, @jasnell I think you two had expressed that the there was an expectation of representation from the Node.js project.  I don't want to be the only person raising this concern if I'm incorrect in thinking we at least have to address the issue in some way.

jasnell:
> I think you two had expressed that the there was an expectation of representation from the Node.js project

Yes, when I did the survey a while back, that was the one thing that has overwhelming consensus.

proposals/stage-0/stage-1/community-board-representation/README.md
MylesBorins: There are two different stages here. The first would be the CPC decided how the seat would be allocated (and likely process for allocation. Then following through on that process.

The CPC could decide to make the seat voted on by the entire membership, they could also decide to give the seat directly to the Node.js project to design a process around, or alternatively could assign it to a specific person.

MylesBorins:
In the last meeting the floating seats were discussed without objection, although we did not discuss the additional seat for the individual membership.

There is no reason why the first iteration of the CPC couldn't assign the seat to the Node.js project. FWIW in the original survey that james implemented the idea of floating seats. The question was a bit leading in my opinion... it was specifically "The Node.js project must have representation on the board". I would be curious how the results would change with the additional context people have now and with additional models for board representation.

proposals/stage-0/stage-1/community-board-representation/README.md
hackygolucky: My assumption of this proposal was that Node.js is separate. We haven't solidified or separated the idea of whether all constellation projects get board seats or JUST Node.js, but I think that the Node.js seat is very separate from this overall.

hackygolucky:
My assumption when talking with Myles about this proposal was that the Node.js dedicated Board seat is a separate concern than the CPC/Community Board seats. We haven't solidified or separated the idea of whether all constellation projects get board seats or JUST Node.js, but I think that the Node.js seat is very separate from this overall.

jasnell:
I'm not really sure how the question was leading as it was clearly presented as a statement for which respondents were asked to clearly indicate agreement or disagreement. It was the only question in the survey where the results were almost exclusively one sided. My read on the results is that respondents very clearly believe that the Node.js project should have representation on the board. However, as with everything, it's always good to gauge consensus again as discussions evolve.

mhdawson:
@hackygolucky, I'm pretty sure at one point in the discussion Sarah mentioned that the seat that Kubernetes has is not fixed in the charter, instead, it is more along the lines of what I had written out in #70.

In the case where a seat for constellation projects is separate, then this proposal is incomplete as it would need to cover that case as well (since as I understand it, it attempts to cover the issue of board representation more broadly than I had in #70).

proposals/stage-0/stage-1/community-board-representation/README.md
mhdawson: @MylesBorins I agree with what you say, but when I said
> We could just have a vote from all members and since Node.js has more members then it would be able to select a representative. 

I meant all members of the combined projects within the Foundation, as that was how the process was outlined for voting in the initial proposal.

hackygolucky:
If we only get 3 community seats total, whether it be CPC member, Individual Membership Director, CPC allocated Director, I think that makes sense. I don't think it makes sense to take one of those 3 seats and just say it's automatically going to Node.js. I don't understand what outcome that gets us to that is more beneficial than it being a flexible seat for of-the-times circumstances.  If we expect a Director to represent the needs of the groups that have put them in that seat, that seems like a very outsized representation for us that the person would just represent Node.js as a Node.js dedicated seat vs. being a Node.js TLC member(just an example) who happened to get in one of those 3 spots and they know both the context of Node.js's needs but aren't bound solely by it. From an expectations management perspective, I think this could be really challenging. I'm not a no on any of this stuff but I'm just trying to think through all we've seen happen over the past few years and what helps here...All thoughts here. Happy to hear other perspectives.

mhdawson:
>I don't think it makes sense to take one of those 3 seats and just say it's automatically going to Node.js.

The language in #70 is that the CPC decides how they are allocated each year.  After the first year they can be allocated however the CPC sees fit.  The inclusion of the suggested allocation for the first year is an attempt to address the expectations we've heard from the Node.js project and to follow a pattern that was successful for CNCF in that respect.  It seems like a reasonable compromise to me.  But agree we need more people chiming in on whether it's reasonable and how we might handle any expectations from the Node.js community.

mhdawson:
Personally, I'd really like us to have 4 (1 CPC representative, 2 as outlined in #70) and a 4th that activates once we have hit the Individual membership threshold.  In the early discussions, the message I heard was that having a larger board was not necessarily an issue.

proposals/stage-0/stage-1/community-board-representation/README.md
MylesBorins: @mhdawson when I had initially proposed the floating board seat I never intended for the allocation to be voted on by all members. Can you point me to the proposal where that was outlined

MylesBorins:
@mhdawson I think that it will be tough for us to add an extra community seat to the board. I also think, FWIW, that is a lot of volunteer labor to be expecting. The community board representation take quite a bit of time... and I'm unsure that we would have 4 qualified candidates at any time to do the work.

Happy to talk to folks on the board and get a pulse on it... but my initial gut reaction is that adding a 4th seat is not a good idea.

For this first iteration I'm imagining it would work out to 1 CPC seat, 1 floating seat (likely assigned to Node.js), and waiting for individual membership seat to activate.

What would you imagine would be a could use for the extra seat we are talking about introducing here? Who would likely be filling that vacancy?

proposals/stage-0/stage-1/community-board-representation/README.md
```diff
@@ -0,0 +1,64 @@
...
+  - delegated on an annual basis and approved by Board
+* 1 board seat for the individual membership program
```
jorydotcom: It's hard to +1 this without understanding what the ind. membership program will look like, what kind of benefits we provide individual members vs. corp. members, how they can join, etc. For the record I am +1 to the idea, but I am also aware that this arrangement can create tension if it's not designed correctly (everyone has to have "skin in the game" and everyone needs to understand what it means to serve as the individual director).

proposals/stage-0/stage-1/community-board-representation/README.md
jorydotcom: to me, this could be an interesting way for projects to rotate representation on the board, as their needs and priorities change. If there were a particularly short term limit for this seat, say a year, I could see the seat going to Node for a term or two, then say NodeRed or JerryScript for a term, etc as needed. It would require that these projects are actively participating in the CPC and communicating with their peers about what they're doing. IMO direct access to the board for each project shouldn't need to be a constant thing - just a need when say, the project is trying to achieve greater adoption or corp support

chowdhurian: The size of a project will make it more likely that a board member will be elected from that project but it doesn't guarantee it. Project members can and should vote for a person who can represent their interests, whether they originate in their project or not. Given the scope of the seat's responsibilities, a person should ideally have some exposure beyond their own project.

On the one hand, a board member from Node.js project could make the other projects have too little direct representation. On the other hand, a board member from a small project could have outsize direct representation.

Having the seat be activated by number of members seems a good way to give projects the right incentives.

MylesBorins:
From the committee this morning this can land with a reference to the already existing proposal

PR #82, opened 2019-01-23
Create Individual Membership representation proposal

hackygolucky:
Creates proposal for the Individual Membership Director as representation at the $FOUNDATION Board level. Prior iterations of this have been researched and lessons learned from the Node.js Individual Membership program.

A separate proposal for the Individual Membership program can be found [HERE](https://github.com/nodejs/bootstrap/pull/81).


proposals/stage-0/individual-membership-director/README.md
```diff
@@ -0,0 +1,73 @@
...
+In order for the Individual Membership Director to provide high fidelity context on the Board, this seat on the Board would only exist with a threshold contingency: once the individual membership roster hits 2000 active members, an election would be triggered for representation on the $FOUNDATION Board of Directors. The term of the elected seat would be for one year. When the year terminates, only if the 2000 person threshold is maintained would a new election occur.
+
+The process for updating the Individual Membership Director seat requires addressing:
+
+- Who is eligible to nominate and run for a Directorship role 
+- Who is eligible to vote for candidates for a Directorship role 
+  
+
+The Individual Membership Director candidates are to be nominated by voting members of the top-level committees of projects and the CPC. The active members of the Individual Membership program would then vote via our proposed election system on the candidate of their choosing.
+
```
jorydotcom: Would nominees be limited to people who had paid individual membership dues, and if so, what class of membership do they have to be (any/bronze/silver/etc)?

MylesBorins:
Does it maybe make sense for us to merge this into the [community-representation-proposal](https://github.com/nodejs/bootstrap/tree/master/proposals/stage-1/community-board-representation)? They seem to overlap

proposals/stage-0/individual-membership-director/README.md
hackygolucky: Excellent question. In the past, you did have to be an individual member to be nominated. I think that makes sense and there's a very low hurdle with this new model. I don't think it should be restricted by tier, so I'd say _any_. If that sounds good, I'll clarify in this line.

MylesBorins:
I'd be up for pulling this in as a separate markdown in the other proposal
if you are up for it. Happy to have it here as well

hackygolucky:
@jorydotcom can you review the most recent change and let me know if that will suffice? 

I think after this merge I'll likely try to accelerate this to stage 3 consideration, as we haven't had much further objection and @MylesBorins had requested combining this proposal for discoverability and archiving's sake with the Individual Membership program proposal which is already at stage 3.

proposals/stage-0/individual-membership-director/README.md
```diff
@@ -0,0 +1,73 @@
...
+Stage 0
+
+### Champion
+[Tracy Hinds]()
+
+### Description
+
+Board representation
+
+The Individual Membership program will bring together the ecosystem around $FOUNDATION projects, users who are not current contributors, that we want to feel more engaged and welcoming than other languages, through participation, input, and accessing the benefits of the foundation that have historically been perceived as available only to contributors. This program codifies a relationship with our ecosystem, as an opportunity for member companies to feel more connected with the members of the Individual Membership through systematic communication building.
```
Trott: Optional suggestion: It took me three or four attempts to read the first sentence before I realized that "users who are not current contributors" is not in addition to the ecosystem but part of it. The sentence wasn't making any sense at all. Honestly, I'm still not sure it is? Can we maybe split it into two or more shorter sentences?

proposals/stage-0/individual-membership-director/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+In order for the Individual Membership Director to provide high fidelity context on the Board, this seat on the Board would only exist with a threshold contingency: once the individual membership roster hits 2000 active members, an election would be triggered for representation on the $FOUNDATION Board of Directors. The term of the elected seat would be for one year. When the year terminates, only if the 2000 person threshold is maintained would a new election occur.
```
Trott: Optional suggestion: I'm not sure what "high fidelity context" means here. Is there another perhaps more straightforward phrase we can use?

MylesBorins:
Hey @hackygolucky do you think we should update this to stage 1 and try and get it landed?

proposals/stage-0/individual-membership-director/README.md
```diff
@@ -0,0 +1,73 @@
...
+Given the goals outlined in this proposal,
+
+- NPS score improves by 10% after base is established 
+    - With the introduction of better feedback mechanisms, we should be able to move some detractors to passives, and some passives to promoters. 
+
+- Annual survey from membership reflects confidence in Director as effective 
+
+### Why this proposal is important
+
+With the historical data we have regarding individual membership, if Board Directors wish for the Individual Membership to be representative of the larger Node.js engaged ecosystem, we know that the number and makeup of the prior votes was not statistically significant nor representative. The Individual Membership did not have a clear path, role, or definition.
```
molant: There are a couple call outs to Node.js specifically. Should those be replaced for a term that represents all the projects in the foundation?

ljharb: what other JS platform would we include? All projects in the foundation are JS projects; are there any for which node isn’t relevant?

molant: Browsers comes to mind, but we don't need to call out any specifically 😊
I think "larger JavaScript engaged ecosystem" is more representative, there are many projects in the foundation that do not require Node.JS to run (moment, lodash, amp and I'm sure many more).

ljharb: all of them require node.js to build, and all distribute a node module, so node is relevant to them as well.

Is there any project that is not impacted by node? Is there any other platform that impacts every project?

I suspect the answer to both is "no".

molant: > so node is relevant to them as well.

Yes, but JS is the common denominator and here we are calling out particularly Node.JS and not JS in general. I believe there's a significant difference between needing Node.JS to build (which only applies to the developers) and Node.JS to run or use it (which applies to consumers which is several orders of magnitude bigger).

> to be representative of the larger Node.js engaged ecosystem

This phrase makes me think about pushing just for Node.JS and not JavaScript overall and we are in the OpenJS Foundation.

ljharb: JavaScript is node and browsers, primarily. There’s other engines, and they matter, but it would be shortsighted of the foundation to ignore the dominant server/build platform in the space.

molant: > t would be shortsighted of the foundation to ignore the dominant server/build platform in the space.

I haven't said we should ignore Node.js. What I'm saying is this language targets only Node.js developers, excluding a lot of potentials members. I believe this is exactly the opposite of what we want to do here.

ljharb: Right - I’m asking who are the non-node developers being excluded? Which projects are thus impacted?

I’m not aware of any, but hopefully you can enlighten me :-)

molant: First paragraph of the proposal (emphasis mine):

> The Individual Membership program will bring together the ecosystem around $FOUNDATION projects, **users who are not current contributors**, that we want to feel more engaged and welcoming than **other languages**, through participation, input, and accessing the benefits of the foundation that have historically been perceived as available only to contributors.

It talks about **users** and **other languages**, not **developers** and **Node.js**.

As you know, until relatively recently only browser vendors participated at the W3C even though the community had a lot to say. They were **the users**, but didn't have a say even if they wanted (or there wasn't a clear direct way). This is the same scenario. We are excluding all the users of the projects we are creating.

Need more concrete scenarios?
A WordPress developer uses a few of the foundation projects (e.g.: jQuery, lodash, or the webhint browser extension to improve their website) and wants to get more involved with the foundation. He could very well not be a Node.js developer or not identify as one.

Another example? React Native developers. They use a lot of libraries published to the registry (which is not Node.js). Do all of them consider themselves Node.js developers? I can guarantee you that most of them don't.

I don't know what else I can say to make you see that this wording does not reach every JavaScript user or developer that could become a member, so I will stop indulging you.

ljharb: That's a fair point; although all of those React Native users are *using* node to npm install things, and the _projects_ that Wordpress dev uses were all built with node, the Wordpress dev may never use it.

MylesBorins:
As the individual supporter program is moving forward I'm going to go ahead and close this

Thanks for all the hard work on this @hackygolucky !

PR #92, opened 2019-02-04
proposal: expectations to stage 2

MylesBorins:
I believe that the major bits we didn't have consensus on have been
resolved in other proposals. This PR moves expectations to Stage 2,
so we can hand it off to the boards and hopefully move it to Stage 3
in a timely fashion.

Very minor tweaks from Stage 1, specifically removing language about
bits that we didn't currently have consensus on.

/cc @hackygolucky

### DIFF
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -11,14 +11,13 @@ Anyone may submit an idea for a policy or program following the [staging process
 
 ## Stage 1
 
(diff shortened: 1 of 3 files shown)


MylesBorins:
From today's meeting there were no objections to landing this. Lets give it a week to collect feedback and land in next weeks meeting as stage 2

PR #103, opened 2019-02-17
proposal: community board reps to stage 2

MylesBorins:
Changes based on current discussion.

Refs: https://github.com/nodejs/bootstrap/pull/102/

### DIFF
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -11,12 +11,12 @@ Anyone may submit an idea for a policy or program following the [staging process
 ## Stage 1
 
 * [CPC Charter](./stage-1/CPC_CHARTER)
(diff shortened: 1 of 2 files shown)


MylesBorins:
Tl;DR

* 1 board seat for CPC
* 1 floating board seat
  - process for seat determined annually by CPC
  - requires 2 Platinum board members to activate
* 1 board seat for individual membership program
  - requires 2 platinum board members and 2000 program members to activate

MylesBorins:
As discussed in today's meeting I am going to land this. I followed up independently with both Jory + Michael after the meeting to ensure that we have consensus. We will move forward with this proposal and keep track of individual membership adoption. If after 6 - 12 months we notice that the program is not trending in a direction to likely trigger the seat we can revisit the allocation.

PR #108, opened 2019-02-28
proposal: move expectations to stage 3

MylesBorins:
Pending board approval from both Node.js and JS
Foundatons


MylesBorins:
added @hackygolucky to sign off as champion and @mkdolan to sign off on behalf of foundations

MylesBorins:
From the board... the request is for us to capture this document into our `artifact/` folder and to break out the components that have not yet been adopted into the Charters and Bylaws into new proposals e.g. `The ED Expectations`.

We have approval to do the above as `Stage 3`... we can add this to the agenda of the next bootstrap meeting and can land this if we have consensus

PR #109, opened 2019-02-28
proposal: project progression to stage-3

MylesBorins:
pending board approval


MylesBorins:
Should have sign off of @jorydotcom as champion and @mkdolan as foundation rep

PR #126, opened 2019-03-19
doc: Updated references to foundation

mhdawson:
Update refernces to specify OpenJS Foundation now
that name is official.


keywordnew:
Now that we have a name, reading OpenJS Foundation has become rather repetitive :)

May be worth including the useful legal pattern of writing a sentence around the first instance along the lines of:

_...in the OpenJS Foundation, hereinafter referred to as the "Foundation"._

The scope of a statement like this is limited to each document, so there won't be spillover in the larger corpus of OpenJS Foundation governance.

mhdawson:
@keywordnew +1 on your suggestion and also +1 on doing in a follow on PR.

Issue #152, opened 2019-04-08
Voting CPC Representatives from Projects


jorydotcom:
Creating this issue to capture the appointments and nominees for the CPC per a question from one of our projects!

--- 

Impact Project Representative Appointments (up to 2 per project)
* Node: Matteo Collina (@mcollina ) & Joe Sepi (@joesepi) 
* [jQuery](https://github.com/jquery/jquery/issues/4344): Dave Methvin (@dmethvin) & Timmy Willison (@timmywil)
* Dojo: Matt Gadd ( @matt-gadd) & Dylan Schiemann (@dylans). 
* Appium: Jonah (@Jonahss) & Dan Graham (@dpgraham)
* webpack: Sean Larkin (@TheLarkInn) & Even Stensberg (@evenstensberg)

Growth & At Large Nominees (up to 2)
* Antón Molleda (@molant)
* Kris Borchers (@kborchers)
* Eemeli Aro (@eemeli)
* Chris Hiller (@boneskull)

molant:
webhint: @molant

Jonahss:
Appium representatives will be @dpgraham and myself 👋

christian-bromann:
As mentioned in webdriverio/webdriverio#3824 for WebdriverIO it will be TSC members of the project, depending who is available in US/Europe.

kborchers:
Architect's nominee is @kborchers per https://github.com/arc-repos/architect/issues/349

knolleary:
Node-RED's rep will be @knolleary (https://github.com/node-red/node-red/issues/2129)

mhdawson:
TSC progress on electing one Node.js rep, https://github.com/nodejs/TSC/issues/689 CommComm will kick off process as well.

eemeli:
Messageformat's nominee will be @eemeli (messageformat/messageformat#232).

amiller-gh:
CommComm nominations now open: https://github.com/nodejs/community-committee/issues/470

mhdawson:
@mcollina has been confirmed as the Node.js Voting CPC member.

keywordnew:
Thank you for taking on the work, @mcollina 🙏🏽

keywordnew:
CommComm will announce a CPC rep on 30th April. Progress is on nodejs/community-committee#470.

dylans:
[...]

PR #176, opened 2019-05-13
Proposal for confirming voting cpc reps from At Large & Growth nominees

jorydotcom:
This PR captures the verbal proposal we have discussed in our meeting session about confirming the nominees from issue #152 - it proposes that we accept all 5 nominees as voting representatives for a period of time, allowing us to select an implement an election system, and that we hold an election for the 2 available positions after the period has ended.

### DIFF
--- a/proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
+++ b/proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
@@ -1,0 +1,39 @@
+# Initial CPC Voting Representative Selection
+>  Stage 0
+
+## Champion
+
+Jory Burson (@jorydotcom)
+
+## Description
+
+In issue #152 we have identified 10 representatives from the Impact projects who will be confirmed as voting CPC Representatives. We have also identified 5 nominees for 2 available positions per the [CPC Charter](https://github.com/openjs-foundation/bootstrap/blob/master/CPC-CHARTER.md) from the At Large and Growth Stage projects. Per the charter, the CPC must hold an election. We do not yet have an election system in place.
+
+This document proposes that we accept all 5 of the initial nominees as voting CPC representatives for an interim period of 6 months, by consensus of the Bootstrap team. During this period, the CPC should take the selection and adoption of an election system as a priority. At the end of the term, there shall be a new call for nominees and an election for the 2 available positions as outlined in the charter.
+
+## Required Resources
+
+none
+
+## Who would be responsible?
+
+Bootstrap Committee / CPC
+
+## How would success be measured?
+
+* nominees are all confirmed
+* election tool/process is selected and implemented
+* new election from nominees is held within 6 months
+
+## Why this proposal is important
+
+I believe this proposal is a 'simplest next step' to booting the functionality of the new CPC with voting members. It welcomes and acknowledges all those who have volunteered their time to participate in an official capacity and invites them to the decision-making table. 
+
+## Unresolved Question
+
+Is 6 months the right amount of time?
+Resolving the > 25% from one company issue
+
+## What is necessary to complete this proposal
+
+Consensus of the CPC/Bootstrap committee. Documenting the requirements & timeline.


proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Champion
+
+Jory Burson (@jorydotcom)
+
+## Description
+
+In issue #152 we have identified 10 representatives from the Impact projects who will be confirmed as voting CPC Representatives. We have also identified 5 nominees for 2 available positions per the [CPC Charter](https://github.com/openjs-foundation/bootstrap/blob/master/CPC-CHARTER.md) from the At Large and Growth Stage projects. Per the charter, the CPC must hold an election. We do not yet have an election system in place.
+
+This document proposes that we accept all 5 of the initial nominees as voting CPC representatives for an interim period of 6 months, by consensus of the Bootstrap team. During this period, the CPC should take the selection and adoption of an election system as a priority. At the end of the term, there shall be a new call for nominees and an election for the 2 available positions as outlined in the charter.
```
mcollina: This implies we have phased terms for impact projects and at large representatives. Is this something that we want, or do we prefer to re-elect the whole CPC every year at the same time?

I'm OK in both sense, but it's something we should discuss / I'm happy to land if this is added to the unresolved questions.

eemeli:
This seems like it's adding somewhat unnecessary complexity compared to the earlier suggestion of just accepting 4/5 of the nominees from at-large/growth projects for the council's first term. At least two of those members will need to step down from their positions very soon after starting, and as Matteo mentions, it seems to lead to a staging of the representatives' selection. Would the next at-large/growth projects' representatives then sit for just six months, a year, or a year and a half? Furthermore, six months from now would probably be rather close to the next summit, which may make arrangements for participation challenging.

As an alternative, would it make sense to define a collective shortened end to the council's first session, i.e. holding the elections in very early 2020? That would mean that this supersized council would sit for about nine months, while allowing the next council a decent time before its first summit -- I at least am finding it a little challenging to get myself to Berlin in two weeks.

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
jorydotcom: That's part of the problem I would love the initial CPC group to address, to determine what's fair. I think it makes sense for the Impact projects to decide this for themselves inasmuch as they are running their own process for determining who their rep is. If Node wants to run an election for their CPC rep every year, and Webpack does it every 2 years, that's fine IMO because it supports individual project autonomy. But what about the nominees for growth + at large rep? 

Great question + should be added to the unresolved section.

jorydotcom:
@eemeli we're suggesting the same thing... it was my suggestion in the meeting to just accept the nominated reps, the only thing I'm adding here is a timebox to ensure that isn't a permanent thing, and I'm being explicit that this initial group would need to design the process moving forward (ie, elections are held every year in January using STV, or whatever).

eemeli:
I don't disagree with the idea here; I'm mostly concerned with the timing with respect to the expected fall collab summit. If we follow past precedent and hold it just before [Node+JS Interactive](https://events.linuxfoundation.org/events/nodejs-interactive-2019/), it would be 9-10 December 2019 in Montreal. Six months from today is 14 November, which is less than a month before those dates. If we want the non-impact project representatives to attend, then that timeline doesn't really work.

Hence my suggestion to extend the proposed timeboxing by a few months to avoid these issues, as well as helping to make sure that the initial representatives have sufficient time to make at least some impact.

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
```diff
@@ -0,0 +1,39 @@
+# Initial CPC Voting Representative Selection
+>  Stage 0
```
MylesBorins: This seems to have enough context to be stage 1

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Champion
+
+Jory Burson (@jorydotcom)
+
+## Description
+
+In issue #152 we have identified 10 representatives from the Impact projects who will be confirmed as voting CPC Representatives. We have also identified 5 nominees for 2 available positions per the [CPC Charter](https://github.com/openjs-foundation/bootstrap/blob/master/CPC-CHARTER.md) from the At Large and Growth Stage projects. Per the charter, the CPC must hold an election. We do not yet have an election system in place.
```
MylesBorins:
```suggestion
In issue [#152](https://github.com/openjs-foundation/bootstrap/issues/152) we have identified 10 representatives from the Impact projects who will be confirmed as voting CPC Representatives. We have also identified 5 nominees for 2 available positions per the [CPC Charter](https://github.com/openjs-foundation/bootstrap/blob/master/CPC-CHARTER.md) from the At Large and Growth Stage projects. Per the charter, the CPC must hold an election. We do not yet have an election system in place.

```

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
MylesBorins: Do we need a proposal to document the decision in the bootstrap committee? It seems to me like we would benefit from a proposal as to how the future elections would work instead.

To Matteo's point, I was under the impression that the election process for CPC would be annual. We can change this if we want, but I think the term limits are a good thing.

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
```diff
@@ -0,0 +1,39 @@
...
+* election tool/process is selected and implemented
+* new election from nominees is held within 6 months
+
+## Why this proposal is important
+
+I believe this proposal is a 'simplest next step' to booting the functionality of the new CPC with voting members. It welcomes and acknowledges all those who have volunteered their time to participate in an official capacity and invites them to the decision-making table. 
+
+## Unresolved Question
+
+Is 6 months the right amount of time?
```
MylesBorins: I was assuming all nominees would run for a year

eemeli: I actually think that a different time period would work better for the council's first session, in order to offset the selection by at least a couple of months from next spring's collab summit. Especially for non-impact delegates it may be rather difficult to arrange for availability on short notice, given the unpredictability of the voting process.

mcollina: I think we are talking about two different topics:

1. should the whole CPC has a fixed term
2. do we want the first year term finishing off on April maybe?

proposals/stage-0/INITIAL_CPC_VOTING_REPS.md
mhdawson: I agree that the process was going to be yearly.  If a project reconfirms the same candidate for a subsequent year though that is up to them.

mhdawson:
We've landed the governance for how the next vote will take place through https://github.com/openjs-foundation/cross-project-council/pull/220.

I think the last thing we need to agree/document is the term of the initial representatives.  I've opened this PR capture that: https://github.com/openjs-foundation/cross-project-council/pull/274

mhdawson:
I think we could probably close this now and complete the discussion on the final issues on #274.

MylesBorins:
Closing as per @mhdawson's comment

PR #186, opened 2019-05-27
doc: add additional expecations on CPC members

mhdawson:
Mostly taken from:
https://github.com/nodejs/admin/blob/master/MemberExpectations.md


mhdawson:
@eemeli I just used the same name as the corresponding file in the Node.js repo. What you you think it should be to be consistent.  I'm happy to change if it makes more sense.

eemeli:
All the other docs are IN_CAPITALS.md. I have no personal opinions on the naming pattern, but it should probably be uniform.

As a separate question, shouldn't this too go through the [staging process](../blob/master/STAGING_PROCESS.md), or is this somehow different?

mhdawson:
In terms of the staging process I think that would be more overhead than necessary in many cases.  My opinion is that we should use if for processes/things we think will require more discussion/consensus building.  Having said that happy to follow that if the rest of the members think its the right thing to do.  I'll update to make the file name be in capitals.

mhdawson:
@eemeli pushed commit to fix the file name.

MEMBER_EXPECTATIONS.md
```diff
@@ -0,0 +1,57 @@
...
+should resign their leadership position.  This does not mean that decisions
+cannot be revisited and discussed within the team at a later time.
+
+Members of our leadership groups must also conduct themselves in a
+professional and respectful manner. Some general guidelines include:
+
+- Remediate quickly when you realize you made a mistake. Leaders are human,
+  and they will make mistakes. However they should act swiftly to
+  acknowledge mistakes and correct them.
+- Aim to remediate first and then discuss.  If other members of the
```
sendilkumarn: nit: space

MEMBER_EXPECTATIONS.md
```diff
@@ -0,0 +1,57 @@
...
+contributors are expected to be respectful of these different viewpoints,
+and to work collaboratively with one another in a manner that constructively
+elevates all contributors.
+
+When decisions are made within the established guidelines and policies of the
+OpenJS Foundation projects or groups, those in leadership roles have a
+responsibility to uphold and respect the decision even if they disagree with
+it. This is especially important in external communications, for example in
+social media. Should the member be unwilling or unable to do so, then they
+should resign their leadership position.  This does not mean that decisions
```
sendilkumarn: nit: space

MEMBER_EXPECTATIONS.md
mhdawson: fixed

MEMBER_EXPECTATIONS.md
mhdawson: fixed

MEMBER_EXPECTATIONS.md
```diff
@@ -0,0 +1,57 @@
...
+- Criticize ideas rather than people, discussing any concerns in person
+  whenever possible, and taking responsibility for our statements.
+
+While the guidelines above focus primarily on the spaces where
+we participate in official OpenJS foundation work (GitHub, IRC, meetings,
+conferences), it is important to recognize that the public behavior
+of members also reflects on the OpenJS Foundation projects and groups.
+
+If you're interested in an introduction to diversity, inclusion, and unconscious bias, 
+[try this free training offered by our partners at the Linux Foundation](https://training.linuxfoundation.org/linux-courses/open-source-compliance-courses/inclusive-speaker-orientation).
```
ChALkeR: Seems to be a redirect to https://training.linuxfoundation.org/training/inclusive-speaker-orientation/.

mhdawson:
All comments addressed, will plan to land in next few days unless there are objections before then.

MEMBER_EXPECTATIONS.md
```diff
@@ -0,0 +1,57 @@
...
+  about others in the OpenJS Foundation projects or team in a public sphere.
+- Build trust by keeping your promises.
+- Be the model of accountability and leadership. Provide the example of
+  ownership and stewardship that everyone can follow to success.
+- Commit to ongoing development and learning best practices for governing.
+- Criticize ideas rather than people, discussing any concerns in person
+  whenever possible, and taking responsibility for our statements.
+
+While the guidelines above focus primarily on the spaces where
+we participate in official OpenJS foundation work (GitHub, IRC, meetings,
```
kborchers: Just noticed a small typo, the 'F' in foundation should be capitalized.

mhdawson: @kborchers fixed.

mhdawson:
All comments address, no objections and lots of approvals.  Landing.

PR #203, opened 2019-06-10
doc: fix charter typos

knolleary:
Fixes a handful of typos in the charter. No material change to the content.

### DIFF
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -23,10 +23,10 @@ the OpenJS Foundation charter at any time, though the Board will not interfere
 with day-to-day discussions, votes or meetings
 of the CPC.
 
-## Section 3. Board’s Role in Setting OpenJS Foundation’s Strategic Direction. 
+## Section 3. Board’s Role in Setting OpenJS Foundation’s Strategic Direction.
 
 The Board will set the overall CPC Policy. The policy will describe the
-overarching scope of the OpenJS Foundation initiative, OpenJS Foundation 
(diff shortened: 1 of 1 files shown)


MylesBorins:
While these are simply errata changes they still need to be reviewed by the board

/cc @brianwarner @mkdolan

CPC-CHARTER.md
```diff
@@ -295,7 +295,7 @@ from the OpenJS Foundation board.
 **Project**: a technical collaboration effort that is organized through the
 project mentorship process and approved by the TAC.
 
-**Contributors**: contribute code or other artifacts, but do not have the right
+**Contributors**: contribute code or other artefacts, but do not have the right
```
eemeli: "artifacts" is valid American English, but do we have a preference for British spelling?

mhdawson: A least in the Node.js project the preference is for US English (I think that is even written into the docs).

kborchers:
LGTM - I would also prefer the US English spelling of 'artifacts' though I am also from the US and obviously biased and will not block a merge based on that change. I would ask that we are consistent though and if necessary, document somewhere which version of English we default to for our docs.

CPC-CHARTER.md
knolleary: Ooops - I thought I had successfully ignored all the US/British alternatives and only fixed true typos. This one slipped past. Will revert this one.

knolleary:
'Artifact' now restored.

MylesBorins:
Approved in board meeting

PR #220, opened 2019-06-20
doc: process for electing non-impact voting members

mhdawson:


GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+  * Confirmation that they can devote the required time/effort.
+  * Their employer  (if applicable).  This is needed due to the limitations
+    on maximum employer representation in the CPC voting members.
+
+* The nomination period will be open for 2 weeks at which point the issue will be locked.
+
+* Once the nomination period closes, the projects will be asked to confirm membership
+  as identified by the nominees.
+
+* Each project will nominate two `voters` through a process of their choosing.
```
MylesBorins: Should this be the Voting members of the CPC?

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+* Each voter will be able to vote for as many or as few candidates as they wish.
```
MylesBorins: This over CIVS?

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+* Those candidates who receive the greatest number of votes will be confirmed as
+  the winners. In the case of a tie between 2 or more candidates, there will be coin
+  tosses between those candiates in order to select the winner.
+
+
```
MylesBorins: lots of extra newlines here

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+* The nomination period will be open for 2 weeks at which point the issue will be locked.
+
+* Once the nomination period closes, the projects will be asked to confirm membership
+  as identified by the nominees.
+
+* Each project will nominate two `voters` through a process of their choosing.
+
+* Each voter will be able to vote for as many or as few candidates as they wish.
+
+* Those candidates who receive the greatest number of votes will be confirmed as
```
MylesBorins: should we maybe do a run off election rather than coin toss?

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
 or else the discussion will continue. Simple majority wins. Only Active
 Members participate in a vote.
 
+
```
MylesBorins: extra newline

GOVERNANCE.md
eemeli: No, not the voting members. It would probably be good to qualify this to mean only the non-impact projects:
```suggestion
* Each non-impact project will nominate two `voters` through a process of their choosing.
```

GOVERNANCE.md
eemeli: Definitely.

sendilkumarn: How about using the number of projects they are actively part of as a selector rather than a coin toss?

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+## Electrion process for Non-Impact Voting CPC members
+
+As per the CPC charter, the non-impact projects elect 2 voting CPC members:
```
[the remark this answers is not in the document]
dmethvin:
```suggestion
As per the [CPC charter](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md), the non-impact projects elect 2 voting CPC members:
```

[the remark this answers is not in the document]
dmethvin: Seems like a good idea.

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+* Once the nomination period closes, the projects will be asked to confirm membership
+  as identified by the nominees.
+
+* Each project will nominate two `voters` through a process of their choosing.
+
+* Each voter will be able to vote for as many or as few candidates as they wish.
+
+* Those candidates who receive the greatest number of votes will be confirmed as
+  the winners. In the case of a tie between 2 or more candidates, there will be coin
+  tosses between those candiates in order to select the winner.
```
dmethvin:
```suggestion
  tosses between those candidates in order to select the winner.
```

GOVERNANCE.md
mhdawson: My preference was for being able to vote for one or more without ranking so I started with that one of the 2 options.  Either way I don't think we need to specifically mention the voting tool, just the process that we want to implement.  @openjs-foundation/cpc please chime in on which of the 2 options you prefer.

GOVERNANCE.md
```diff
@@ -65,6 +65,48 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+> up to two Voting members may be nominated by the non Impact projects
+  based on a process set by the CPC.
+
+These members will be elected as follows:
+
+* Once a year an issue will be opened in the CPC announcing that the period for
+  nominations is open.
+
+* Any member from one of the non-impact projects can self-nominate by posting a
```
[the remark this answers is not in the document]
mhdawson: I'm good either way.  @openjs-foundation/cpc please chime in with your preference.

GOVERNANCE.md
mhdawson: We had discussed having multiple rounds but there was some concern over asking people to manage and participate in a number of voting rounds.  @sendilkumarn I'm not sure the number of projects is necessarily a good selector as it gets very subjection to assess the level of participation in a project.

GOVERNANCE.md
mhdawson: Should be fixed now.

GOVERNANCE.md
```diff
@@ -65,6 +65,41 @@ meeting. In both cases the decision must be seconded by a majority of the Team
 or else the discussion will continue. Simple majority wins. Only Active
 Members participate in a vote.
 
+## Election process for Non-Impact Voting CPC members
+
+As per the [CPC charter](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md), the non-impact projects elect 2 voting CPC members:
+
+> up to two Voting members may be nominated by the non Impact projects
```
trivikr:
```suggestion
> up to two Voting members may be nominated by the non-impact projects
```

GOVERNANCE.md
```diff
@@ -65,6 +65,41 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+  nominations is open.
+
+* Any member from one of the non-impact projects can self-nominate by posting a
+  reply which includes the following:
+
+  * Projects they are part of
+  * Description of their background
+  * The reasons they would like to act as the non-impact CPC voting member
+  * Confirmation that they can devote the required time/effort.
+  * Their employer  (if applicable).  This is needed due to the limitations
```
trivikr:
```suggestion
  * Their employer (if applicable). This is needed due to the limitations
```

GOVERNANCE.md
boneskull: given the amount of markdown we’re working with, probably a good idea to add some linting/auto-formatting tooling. github action maybe?

GOVERNANCE.md
boneskull: “two (2)” instead of “2”

😝

GOVERNANCE.md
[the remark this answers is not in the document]
boneskull: yes, email sounds good to me

GOVERNANCE.md
kborchers: I would prefer holding runoff elections in the event of a tie as I'm not too keen on using random selection to choose who represents a potentially large group of projects on the CPC. That said, if the consensus is to avoid multiple rounds of voting, I don't have a better suggestion other than one of the more complex voting systems of which I am also not a big fan.

Jonahss: If the top is a two-way tie, those two should be the two voting members. That makes sense right?
If the second position is a tie, then random, but I suggest giving the candidates the chance to concede to each other first?
If the top position is a three-way tie or more, then maybe we should have a second election?

kborchers: I like that @Jonahss as it should avoid a run off in most cases and only requires a coin toss when necessary rather than defaulting to that.

mhdawson: @Jonahss  if its a 2 way tie I had assumed it would be those 2 based on the existing wording.  Do you think I should clarify that further?

Writing in a chance for consensus first makes sense to me, I'll add that.

I'm neutral on re-vote or random, but you might still end up with three-way or more tie on a revote as well and we'd have what to decide to do in that case.

Jonahss: I assumed that the revote would be limited to just the tied candidates. Is that what you were thinking too?

2 way tie case doesn't need to be clarified I guess.

mhdawson: When I looked at the text again I added some extra clarification, @Jonahss do you think that's better?

GOVERNANCE.md
```diff
@@ -65,6 +65,46 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+As per the [CPC charter](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md), the non-impact projects elect 2 voting CPC members:
+
+> up to two (2) Voting members may be nominated by the non-impact projects
+  based on a process set by the CPC.
+
+These members will be elected as follows:
+
+* Once a year an issue will be opened in the CPC repo announcing that the period for
+  nominations is open.
```
mcollina: we should probably add a reference to how long is the perm of a voting member.

mhdawson:
I think we've reached consensus on this and it is ready to land.  @openjs-foundation/cpc please comment if that is not the case. Otherwise I'll plan to land early next week.

MylesBorins:
landed in 6ab56f4e22

PR #222, opened 2019-06-20
add ljharb to members list

ljharb:
Rebased on top of #219.

PR #224, opened 2019-06-20
Add Sendil to regular members list

sendilkumarn:
Rebased on top of #222

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,13 @@ According to the [CPC Charter](https://github.com/openjs-foundation/cross-projec
 - **Node-RED**: Nick O'Leary (@knolleary)
 - **webhint**: Antón Molleda (@molant)
 
-<!-- ### Regular Members -->
+### Regular Members
+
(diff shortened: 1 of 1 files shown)


mhdawson:
@openjs-foundation/cpc could we get a few more approvals so we can land?

mhdawson:
@sendilkumarn we talked a bit about the process for adding regular members in the last meeting and part of the was to ask what project people are part of so that we can provide an FYI to the project that a member has joined. 

Can you add a comment with the Foundation projects that you are involved in?

sendilkumarn:
> Can you add a comment with the Foundation projects that you are involved in?

mhdawson:
Lots of approvals and no objects >72 hours.  There was some confusion  on whether I should have landed the other regular member PR's but since we've not reverted those I think we should land this one as well.  @openjs-foundation/cpc please comment if you disagree, otherwise I'll plan to land mid next week.

MylesBorins:
landed in c946c85

PR #225, opened 2019-06-23
doc: remove unexpected open paren in MEMBER_EXPECTATIONS

Krinkle:


MEMBER_EXPECTATIONS.md
```diff
@@ -1,7 +1,7 @@
 All participants in the OpenJS Foundation projects and groups must follow the
 [Code of Conduct](CODE_OF_CONDUCT.md). There are further expectations for
 members of leadership teams including the
-([CPC](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md).
+([CPC](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md)).
```
Trott: I think the correct thing to do is to remove the opening parenthesis rather than adding a closed one. It should be `including the CPC` and not `including the (CPC)`.

Krinkle: Ha, indeed. Should've read the full sentence first :). Fixed!

mhdawson:
Lots of approvals, no objections, well past 72 hours landing.

PR #232, opened 2019-06-24
governance: add guests to charter

MylesBorins:
Hey All,

This came up in putting together the work for the project progression process... how do we want to handle invited guests who do not participate in consensus seeking.

I also noticed that we never identify in the charter or governance what qualifies something as an observer. I have similarly not documented how to invite a guest. It seems like perhaps we want a bit more policy around this, but I'm unsure if it belongs in the charter or in the GOVERNANCE.md

We might want to refactor this quite a bit

### DIFF
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -44,7 +44,7 @@ direction of the policies set by the Board.
 
 ## Section 4. Establishment of the CPC.
 
-The CPC is an inclusive group with f types of participants:
+The CPC is an inclusive group with four types of participants:
 
 * Observers
 * Regular members
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -44,12 +44,17 @@ direction of the policies set by the Board.
 
 ## Section 4. Establishment of the CPC.
 
-The CPC is an inclusive group with three types of participants:
+The CPC is an inclusive group with f types of participants:
 
 * Observers
 * Regular members
 * Voting members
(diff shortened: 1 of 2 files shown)


CPC-CHARTER.md
```diff
@@ -44,12 +44,17 @@ direction of the policies set by the Board.
 
 ## Section 4. Establishment of the CPC.
 
-The CPC is an inclusive group with three types of participants:
+The CPC is an inclusive group with f types of participants:
```
trivikr: four?
```suggestion
The CPC is an inclusive group with four types of participants:
```

mhdawson:
As it is currently written

> Observers are free to attend meetings and participate in the work of the
CPC as well as the consensus seeking process.

This says anybody is invited to attend the meetings and they can participate in the discussions and consensus seeking process.  If they object, continue to participate in the discussion and the objection cannot be addressed then the voting members would have to vote.

If the intention was to limit who could be observers we should document what the limitations one how an individual can be an observer (My current understanding was that since we'd not documented any, that anybody was welcome as an observer) that really needs to be added to this PR.  Adding the invited guest category does not make sense unless there are limits on who can show up as an observer.

MylesBorins:
@mhdawson I've opened https://github.com/openjs-foundation/cross-project-council/issues/234 to discuss "Who can be an observer". I'll update this and other PRs based on where that discussion goes.

Thanks for pointing that out!

PR #238, opened 2019-06-26
doc: update governance related to charter changes

mhdawson:
Update to add goverance related to how charters and
charter changes are approved.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -67,7 +67,7 @@ Members participate in a vote.
 
 ## Aproving Project Charters
 
-As per the CPC charter the voting CPC members are responsible for approving project
+Per the [CPC charter section 5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc), the voting CPC members are responsible for approving project
 charters and changes to them as per section
 [5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc).
 
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -68,7 +68,7 @@ Members participate in a vote.
 ## Aproving Project Charters
 
 Per the [CPC charter section 5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc), the voting CPC members are responsible for approving project
-charters and changes to them as per section
+charters and changes to them.
 [5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc).
 
 Approval of the initial charter or changes to an existing charter will
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -79,7 +79,7 @@ conditions:
 
 * there are no outstanding objections
 * there are two or more approvals by voting CPC members
-* the board has been consulted in the case of substantive changes
+* the board has been consulted in the case of substantial changes
 * The PR has been open for at least 14 days
 
 ## Merging PRs into this Repository
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
 or else the discussion will continue. Simple majority wins. Only Active
 Members participate in a vote.
 
+## Aproving Project Charters
+
+As per the CPC charter the voting CPC members are responsible for approving project
+charters and changes to them as per section
+[5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc).
+
+Approval of the initial charter or changes to an existing charter will
+be approved by opening an issue requesting approvle in the cross-project-council
+[repository](https://github.com/openjs-foundation/cross-project-council/)
+
+The request will be approved through our standard concensus seeking process with the following
+conditions:
+
+* there are no outstanding objections
+* there are two or more approvals by `voting` CPC members
+* the board has been consulted in the case of substantive changes
+* The PR has been open for at least 14 days
+
 ## Merging PRs into this Repository
 
 Pull requests that do not change the charter or governance of the CPC can be merged into
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -75,7 +75,6 @@ be requested by opening an issue requesting approval in the cross-project-counci
 [repository](https://github.com/openjs-foundation/cross-project-council/)
 
 The request is approved when:
-conditions:
 
 * there are no outstanding objections
 * there are two or more approvals by voting CPC members
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -80,7 +80,7 @@ conditions:
 * there are no outstanding objections
 * there are two or more approvals by voting CPC members
 * the board has been consulted in the case of substantial changes
-* The PR has been open for at least 14 days
+* The issue has been open for at least 14 days
 
 ## Merging PRs into this Repository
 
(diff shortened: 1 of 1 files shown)


GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+Approval of the initial charter or changes to an existing charter will
+be approved by opening an issue requesting approvle in the cross-project-council
+[repository](https://github.com/openjs-foundation/cross-project-council/)
+
+The request will be approved through our standard concensus seeking process with the following
+conditions:
+
+* there are no outstanding objections
+* there are two or more approvals by `voting` CPC members
```
mhdawson: It has been discussed that maybe this needs to be approved by a quorum of voting members instead?  @openjs-foundation/cpc please chime in.

knolleary: Without refering to the specifics of this case, my gut instinct is there should be as few permutations as possible for how things are approved. That will remove confusion of what is required for any particular decision as they come up.

'Approval by a quorum of CPC voting members' is the standard baseline of approval and I don't necessarily see why this would be different.

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
 or else the discussion will continue. Simple majority wins. Only Active
 Members participate in a vote.
 
+## Aproving Project Charters
+
+As per the CPC charter the voting CPC members are responsible for approving project
```
Trott:
```suggestion
Per the CPC charter, the voting CPC members are responsible for approving project
```

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+charters and changes to them as per section
```
Trott:
```suggestion
charters and changes to them.
```

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+[5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc).
```
Trott:
```suggestion

```

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+Approval of the initial charter or changes to an existing charter will
+be approved by opening an issue requesting approvle in the cross-project-council
```
Trott: > be approved by aopening an issue requesting approvle

Wha?

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+[repository](https://github.com/openjs-foundation/cross-project-council/)
+
+The request will be approved through our standard concensus seeking process with the following
```
Trott: Since `standard consensus-seeking` doesn't add anything that isn't in the bullet points, it's better to say:

```suggestion
The request is approved when:
```

That will prevent people from taking time hunting for non-existent additional criteria in the consensus-seeking process.

GOVERNANCE.md
Trott:
```suggestion
* there are two or more approvals by voting CPC members
```

timmywil: I agree that a quorum of voting members is the goal, but I think this is meant to address how that approval carries over to GitHub. Specifically, forcing a quorum of GitHub approvals on all PRs could get tedious in practice and end up with lots of "please add approval so we can merge" comments. Alongside the other requirements (being open 14 days, having no objections, and the board having been consulted for substantive issues), 2 explicit GitHub approvals is perhaps enough to assume a quorum.

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+conditions:
+
+* there are no outstanding objections
+* there are two or more approvals by `voting` CPC members
+* the board has been consulted in the case of substantive changes
```
Trott: A substantial change would be an important/significant change. A substantive change could be a small/insignificant one, as long as it is a change in meaning. Do we mean that the board needs to be consulted on any change in meaning, even insignificant ones? Or do we mean that the board is consulted for any significant changes?

```suggestion
* the board has been consulted in the case of substantial changes
```

GOVERNANCE.md
```diff
@@ -65,6 +65,24 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+* The PR has been open for at least 14 days
```
Trott: Above it's an issue in the tracker, but here it's a PR. (I'd use "pull request" rather than PR, FWIW...er, for what it's worth.) Decide which one it should be and stick with that one only.

mhdawson: Cut and past issue here, fixed.

mhdawson:
@knolleary the standard approval process for the CPC is through `consensus seeking` not 'Approval by a quorum of CPC voting members'. We can make quorum required  in some cases, but by default it is consensus seeking as per the charter section https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-10-voting

GOVERNANCE.md
```diff
@@ -65,6 +65,23 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+Per the [CPC charter section 5](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md#section-5-responsibilities-and-expectations-of-the-cpc), the voting CPC members are responsible for approving project
+charters and changes to them.
+
+Approval of the initial charter or changes to an existing charter will
+be requested by opening an issue requesting approval in the cross-project-council
+[repository](https://github.com/openjs-foundation/cross-project-council/)
+
+The request is approved when:
+conditions:
```
MylesBorins: I think you want to remove this line

GOVERNANCE.md
```diff
@@ -65,6 +65,23 @@ meeting. In both cases the decision must be seconded by a majority of the Team
...
+
+* there are no outstanding objections
+* there are two or more approvals by voting CPC members
```
MylesBorins: I have mixed feelings about this tbqh. On one hand this will be good in combination with the 14 day rule.. that being said if we have a meeting with quorum of CPC voting members and no objections should it have to wait?

mhdawson: I still come back to "how do you know there are no objections".  Since potentially half the of the members have not seen the PR you don't know.  Yes the 1/2+1 can always override the other 1/2-1 but since we are consensus seeking we should be looking to address objects if there are any, not just ignore them because we have 1/2+1 who agree.  

In my opinion, if in the end if we just say that 1/2+1 can make the decision without giving the rest of the members the option to object and looking to address those objections, we are not really following consensus seeking.

mhdawson: I'll also add that it's ok if we think that some decisions should be majority rule versus consensus seeking but I think that should be explicitly stated that we've made that decision versus it being consensus seeking unless 1/2+1 show up at a meeting and agree.

dmethvin:
Sorry! Pressed the wrong button an there is no undo

dmethvin:
I have substantive input but I'll wait until I'm at my computer tonight.

GOVERNANCE.md
```diff
@@ -65,6 +65,22 @@ meeting. In both cases the decision must be seconded by a majority of the Team
 or else the discussion will continue. Simple majority wins. Only Active
 Members participate in a vote.
 
+## Aproving Project Charters
```
dmethvin: Approving

GOVERNANCE.md
dmethvin: After reading through all the text in context I think I'm okay with it. If we are talking about shortening the time then a solution like a super-majority might be useful, but here it seems like we're trying to address the situation where apathy or the lack of controversy has caused people to ignore the issues. I think we owe it to projects to give them autonomy over their charters unless there are objections within a reasonable period.

GOVERNANCE.md
mhdawson: fixed.

PR #218, opened 2019-06-20
enh: add project links

evenstensberg:
Closes https://github.com/openjs-foundation/cross-project-council/issues/215 and fixes some whitespace

PR #247, opened 2019-07-02
doc: add https for esprima website

trivikr:
https link wasn't added in #218 as their ssl certificate was invalid
it was followed up and fixed in https://github.com/jquery/esprima/issues/1975


kborchers:
Open 72 hours, multiple CPC member approvals, no objections, merging

PR #259, opened 2019-07-17
Add Tierney Cyren (@bnb) as a regular member

bnb:
Adds myself as a regular member ❤️


MylesBorins:
We likely want to draft better governance for onboarding (and offboarding) regular members.

* How long does this PR need to be open?
* What GitHub orgs / teams do they need to be added to?
* What mailing lists do they need to be added to?
* How can someone object? Publicly or Privately?

Likely this should be broken into another issue to track.

mhdawson:
@bnb since I know I'll add that you are active in the Node.js project and I assume that is the project that you would reference to meet the 3 month activity requirement in of the member projects.  Making it explicit since that is one of the requirements we should be checking.

Issue #214, opened 2019-06-19
Decide on a timeline for selecting/electing/ratifying the next CPC members


eemeli:
The inaugural meeting of the first CPC was held on 20 May 2019, ten days before the first OpenJS Collaborator Summit in Berlin. Here's what our current charter says:
> Voting members serve for a term of 1 year and must be re-nominated and ratified by the Voting CPC members each year.

We should figure out when in particular the elections of the next non-impact voting members will be held, as the likely proximity of the end of the nominal one-year term to next spring's OpenJS Collab Summit (should one be held) may make travel arrangements difficult for elected members. This term limit and its proximity to the summit may be challenging for the selection of other voting members as well.

As we've already flexed on the number of non-impact voting members, I believe there's some precedent here for flexing on the understanding of "1 year" for this initial CPC. I would therefore propose that a determination be made that this "year" ends either at least a month before or a month after a presumed spring 2020 OpenJS Collab Summit, and then to work backwards from there to develop a more complete timeline.

Development of the specific election method is being tracked by #161. As we do not yet have any voting members nominated by regular members, I believe the timeline of their selection is a separate issue from this one.

I myself am a non-impact voting member and so a rather interested party in this issue. My main interest here is to establish clarity on the timing sufficiently well ahead of it becoming potentially emotionally laden for any parties.

eemeli:
Posting this intentionally as a comment: I think the best solution would be to end the first "year" 6-8 weeks before next spring's summit, so that any newly selected voting members may have sufficient time to make travel arrangements (if possible) in order to participate in the summit. Meeting people face-to-face always makes it much easier to continue discussions with them later online.

mhdawson:
I agree having the voting members elected in advance is a good idea.  My suggestion is that we make the current term run until the end of April.  We don't really know when the summit will be held, but I think at least for this year, April is likely going to before the next one after Montreal.

mcollina:
I think we should:

1. start the election process  at the beginning of March
2. elect the new representative by April 2nd
3. the new representative starts on May 1st

Essentially there would be one month of handover, and travel plans could still be feasible.

dmethvin:
Seems like a reasonable plan, assuming the collab summit stays in a narrow range of dates.

mhdawson:
+1 to @mcollina's suggestion.

keywordnew:
+1 @mcollina's pre-summit timeline seems reasonable to me

eemeli:
[...]

PR #274, opened 2019-07-30
doc: add gov for first term of voting members

mhdawson:
Refs: https://github.com/openjs-foundation/cross-project-council/pull/176

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -74,8 +74,8 @@ From: https://github.com/openjs-foundation/cross-project-council/blob/master/CPC
   by the Voting CPC members each year.
 
 Due to the bootstrap process, the term for the first year will be
-shortened such that the initial term for all voting members ends April 30th, 2020.
-Subsequent terms wil be for a period of 1 year ending the following April 30th.
+shortened such that the initial term for all voting members ends March 31st, 2020.
+Subsequent terms wil be for a period of 1 year ending the following March 31st.
 
 ## Election process for Non-Impact Voting CPC members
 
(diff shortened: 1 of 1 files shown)


eemeli:
Other relevant issue: #214

mhdawson:
Happy to be more conservative and change to march 31

mhdawson:
@eemeli updated.

kborchers:
I don't think setting voting member terms based on when a conference takes place is a good idea. I think dates should be set and then if the transition can happen before an event where a meeting will take place, that's great. Otherwise, elections should still happen and the CPC continues to function without being beholden to some conference date that can and will likely change due to many factors out of our control.

eemeli:
Just to clarify, I do agree that we should not be beholden to the whims and vagaries of inconstant conference dates. But I do think that we should not make it more difficult to organise a potential European collab summit next year, or in subsequent years.

My point isn't about the timing of any specific conference, but about the fact that there aren't many in late May or June while there are plenty in April and early May. Next year is not really special in this regard. So, if the CPC terms ends on 31 March, it's much easier to organise a Spring collab summit. If it ends on 30 April, it's more difficult.

kborchers:
I should also clarify that I am fine with the March 31 date as well so I think we are on the same page with that. :)

FWIW, to travel to Europe for an event in the spring, I would likely need to know it was happening by the beginning of the year to be able to plan from both work and personal schedule perspectives. Not sure if others need as much lead time but in general, I usually can't get approvals, etc with only a month or two especially since I am not filling this role as part of my job so I either have to find a way to tie work-related travel to the trip or take vacation time.

mhdawson:
Agreed we could land this in the CPC meeting today. Landing

PR #278, opened 2019-08-02
Travel Fund - Promotion to Stage-1

Jonahss:
Second stage in a 4 stage process for a proposal.

This proposal is to move governance of the Foundation Travel Fund to the Cross Project Council.

In the last PR, we were just setting up the general purpose of the proposal. In this PR, we can focus on the details of how the fund should be managed. I've included the relevant files from the nodejs/admin repo in this PR, and have made some proposed changes already. Since the files are added in this PR, here is a better diff for viewing the changes I've made:
https://github.com/openjs-foundation/cross-project-council/compare/c9332b8...Jonahss:travel-fund?expand=1

Ideally, I'd like to keep the process as similar to the existing process as possible, but some changes have already been suggested and we should discuss those, as well as budgeting safeguards.

In order to inform the discussion, I'm working on surveying the projects for how they anticipate how they may use the travel fund in the upcoming year. I will also provide some analysis of the past few years of fund usage.

PR for previous stage: https://github.com/openjs-foundation/cross-project-council/pull/268

Addresses #172

@mentions for those who had comments on the previous PR which we said we'd discuss here:
@mcollina, @joyeecheung, @bnb,

Issue #172, opened 2019-05-08
New process for CPC reps to share responsibility for Travel Fund


keywordnew:
We'll be needing a process for the CPC to begin voting on [Travel Fund](https://github.com/nodejs/admin/blob/master/MEMBER_TRAVEL_FUND.md) PRs together with the @nodejs/tsc and @nodejs/community-committee in the near future.

This is based on the interpretation of the Travel Fund being on the CPC charter as a shared resource it's responsible for ensuring. Possibly a candidate for [Steps to Bootstrap CPC](https://github.com/openjs-foundation/bootstrap/projects/2).

Trott:
I'd recommend being very careful in the approach taken here. The travel fund was set up by the Node.js Foundation board to promote Node.js and to be administered by the Node.js TSC and CommComm. Basically negating that and making it an OpenJS thing would have the appearance of confirming the fears of some on the TSC that the foundation merger would dilute Node.js resources.

Here's a way to get to the same place, probably, but with a process that (hopefully) doesn't provide fodder for anyone who might want to portray this as the merger taking something away from Node.js: 

* Leave the current travel fund alone for the rest of the year. (There might not much money left in it at this point anyway, once you deduct all the requests that have already been made. Maybe someone on the Foundation side can report the current projected balance.)

* Create a new fund for the CPC. It can be small this year. CPC (and Node.js for that matter) can always ask for more money if it runs out.

* Next year, dissolve the Node.js fund, but let the TSC and CommComm know that if they have a legitimate reason to want their own fund, they can request it.

Ultimately, this has the same effect as just having the CPC start voting on Node.js Travel Fund requests this year, but the way we get there matters (at least in my opinion).

Trott:
(Actually, the above arguably has a better result in that travel fund ultimately is a CPC thing and the Node.js TSC and CommComm don't have to co-administer it.)

MylesBorins:
Do we need a foundation wide fund? This seems to me like something that should perhaps be a project based budget.

mcollina:
I have a _slightly_ different proposal, essentially splitting the reason why people use the Node.js Travel fund right now:

1. Attending the Collab Summit
2. Promoting Node.js at other events
3. Diagnostics summit and similar

In 2020, I would like to see:

1. a fund for the Collab Summits, managed by the Collab Summit team
2. a (micro) fund for the CPC - in case there is some travel planned (might not be needed, or handled in an ad-hoc form)
[...]

PR #283, opened 2019-08-08
Travel Fund - stage 2

Jonahss:
Continuing to move this forward.

PR for previous stage: https://github.com/openjs-foundation/cross-project-council/pull/278

Addresses #172


mcollina:
I see no differences with the Stage-1. I don't think any are needed, but I'm asking just in case.

I'm not sure if we would like to land this _after_ the Node.js TSC and CommComm has approved the move. Anyway, I think that has to happen before it goes to the Board to increase the size of the travel fund.

MylesBorins:
Do the tsc and commcomm need to explicitly approve?

Since everyone is able to participate in this repo would the lack of
objections not be implicit approval? In which case in might make sense to
set a time box, such as 2 weeks, for comment?

Jonahss:
Nothing different from stage-1 in here yet, yeah.

I think board approval is part of the step from stage 2 to stage 3. 
Yeah, the conversation in https://github.com/nodejs/admin/issues/398 is currently signaling approval of this operation. Plenty of mentions and reviews, some support. No concerns or objections.

I guess what we should do for this PR is flush out the few TODOs I've got in there. I can also continue research on expected future usage and suggest a budget increase to accompany this proposal. So far, projects which have contacted me have no plans to use the travel fund, or would like to use it to have a few members attend the collaborator summit, as happened earlier this year.

mhdawson:
@MylesBorins I don't think a explicit approval from the TSC or CommComm is necessarily required, however, I also don't think in this case assuming those groups track issues outside the Node.js project is good enough either.  What I'd like isconfirmation from our CPC reps (Joe and Matteo) that the issue has been discussed at the TSC and CommComm level and that either:

* there is general support/agreement 
* there is a list of concerns that should be brought back to the CPC for discussion before it proceeds.

@joesepi, @mcollina does this make sense to you?

MylesBorins:
The `Unresolved Questions` and `What is necessary to complete this proposal` sections should likely be updated for this to go stage 2... although that isn't a blocker for this PR

mhdawson:
@mcollina since I had to edit my comment above.

Jonahss:
Sorry I haven't been more active on this, am currently on vacation. Will be back next week :)

MylesBorins:
This has been open for 15 days with no objections. I think we should land this after the next CPC meeting if there are no objections

PR #284, opened 2019-08-09
Update Interledger website

bnb:
Brings the Interledger link in-line with other links – a website, rather than the GitHub org (removed link redirects to GitHub org)


kborchers:
I'm not too concerned with this change but just so everyone is aware, interledger.org is the site for the Interledger protocol. The OpenJSF project is Interledger.js which is the JavaScript implementation of that protocol. Similarly, [Hyperledger Quilt](https://www.hyperledger.org/projects/quilt) is the Java implementation. That distinction isn't super clear on that home page.

PR #294, opened 2019-08-15
doc: rename acceptance criteria to graduation criteria

tobie:
The acceptance criteria listed in the Incubation phase are actually exit criteria.

Hence adding an "Exit Criteria" subsection and moving them there.

### DIFF
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -153,7 +153,7 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
-1. 2/3 vote from the CPC to accept project (during silent period).
+1. consensus within private mailing list to move into incubation process
 
 #### Graduation Criteria
 
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -153,9 +153,9 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
-1. 2/3 vote from the CPC to accept project during silent phase
+1. 2/3 vote from the CPC to accept project (during silent period).
 
-#### Exit Criteria
+#### Graduation Criteria
 
 1. Completion of the on-boarding checklist
 1. 2/3 vote from the CPC to accept project
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -153,7 +153,7 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
-1. 2/3 cote from the CPC to accept project during silent phase
+1. 2/3 vote from the CPC to accept project during silent phase
 
 #### Exit Criteria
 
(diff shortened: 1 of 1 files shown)


PROJECT_PROGRESSION.md
```diff
@@ -153,6 +153,10 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
+1. 2/3 cote from the CPC to accept project during silent phase
+
+#### Exit Criteria
+
 1. Completion of the on-boarding checklist
 1. 2/3 cote from the CPC to accept project
```
ljharb:
```suggestion
1. 2/3 vote from the CPC to accept project
```

PROJECT_PROGRESSION.md
```diff
@@ -153,8 +153,12 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
+1. 2/3 cote from the CPC to accept project during silent phase
```
tobie:
```suggestion
1. 2/3 vote from the CPC to accept project during silent phase
```

mcollina:
I do not think the term “exit” is clear enough in this context. A project can “exit” the foundation, while here I think you mean it will exit the incubating stage. I think the previous wording was less ambiguous, but as you mentioned not 100% correct. Maybe we can clarify better?

tobie:
Graduation criteria?

PROJECT_PROGRESSION.md
```diff
@@ -153,8 +153,12 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
+1. 2/3 vote from the CPC to accept project (during silent period).
```
MylesBorins: I'm not 100% sure that this is accurate. My intention when drafting this policy would be that there was consensus to move a project into the incubation but not a formal vote. That would only happen at the end of incubation to officially add project to foundation. TBH I would prefer we didn't even have a vote then... but I was under the impression that we needed some official vote during the process at some point.

Suggested alternative language... don't think it's great. Please feel free to replace with something else that captures the spirit.

```suggestion
1. consensus within private mailing list to move into incubation process
```

mhdawson: +1 as my understanding is that is was more of a "no objections" in the silent period versus a formal vote.

PROJECT_PROGRESSION.md
```diff
@@ -153,8 +153,12 @@ It is expected that incubating project will make an active effort to work throug
 
 **Acceptance Criteria**
 
+1. consensus within private mailing list to move into incubation process
+
+#### Graduation Criteria
```
eemeli:
```suggestion
**Graduation Criteria**
```

All of the other sub-titles are using `**bold**` rather than `#### h4` styling. This should match the rest. Alternatively, the styling of the other titles should be fixed.

ljharb: Heading styling creates permalinks, so it’s a better choice.

tobie: That’s been fixed for the rest of the page with https://github.com/openjs-foundation/cross-project-council/pull/293 which has already been merged.

MylesBorins:
Multiple approvals and open for 14 days, landing

Issue #288, opened 2019-08-13
Duplicate onboarding checklist


tobie:
There's an onboarding checklist at the bottom of [NEW_PROJECT_APPLICATION.md](https://github.com/openjs-foundation/cross-project-council/blob/master/NEW_PROJECT_APPLICATION.md#onboarding-checklist) and one as an [issue template](https://github.com/openjs-foundation/cross-project-council/blob/master/.github/ISSUE_TEMPLATE/project-onboarding-checklist-template.md). Would you take a PR that referenced the latter in NEW_PROJECT_APPLICATION.md instead of keeping two copies around which _will_ get out of sync?

jorydotcom:
@tobie At the moment I'm glad to have the checklist included in the issue template because we're also putting existing projects through the process and it's helpful to see the status of the checklist for a given project right there on the issue. But we can def. discuss at the CPC meeting today & see if there are some other optimizations / adjustments to make. I don't think the issue template is necessarily the right format long term, for example

tobie:
Fair enough, though I think not being sure _where_ the checklist ends-up (and being open to moving it at a later stage) is a different concern than avoiding to create two, potentially out of sync checklists.

tobie:
Lurking on the conf call, I think there is a misunderstanding between us, @jorydotcom.My concern is about a duplicate checklist between the NEW_PROJECT_APPLICATION.md file and the ISSUE_TEMPLATE/project-onboarding-checklist-template.md file. Adding the checklist to each issue created is perfectly fine. It's the whole point!

MylesBorins:
I've opened a PR that should remove the duplication

PR #304, opened 2019-08-19
doc: remove duplicate checklist

MylesBorins:
This removes the duplicated checklist in the issue template and
instead links to the checklist in the new project application

Fixes: #288

### DIFF
--- a/.github/ISSUE_TEMPLATE/project-onboarding-checklist-template.md
+++ b/.github/ISSUE_TEMPLATE/project-onboarding-checklist-template.md
@@ -11,22 +11,4 @@ assignees: ''
 
 **Stage: ${Project Stage}
 
-- [ ] Adopt the [OpenJS Foundation Code of Conduct](https://github.com/openjs-foundation/cross-project-council/blob/master/CODE_OF_CONDUCT.md) - update the project's Code of Conduct file
-- [ ] Update project CoC reporting methods to include OpenJS Foundation escalation path
-- [ ] Transfer official domains to OpenJS Foundation
-- [ ] Identify and document other core project infrastructure, including hosting, CDNs, CI/CD tooling, etc. If the project anticipates anticipate additional needs, document that as well. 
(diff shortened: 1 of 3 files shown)


MylesBorins:
@jorydotcom I know you had mixed feelings about landing this change. Happy to close it just lmk

tobie:
Curious why you don't do that the other way round (i.e., keep the checklist in the issue template).

MylesBorins:
@tobie the issue template is in a hidden directory `.github` and not super discoverable for folks aside from when the issue is created.

tobie:
Fair enough.

kborchers:
Hmm, I definitely agree with removing the duplication but I think it makes more sense for the list to live in the template where it will actually be used and then we can link to the template from the application for folks to review it when reviewing the application process. Asking folks to copy/paste when creating the application issue sort of defeats the purpose of the template IMO.

If others disagree and we decide to merge this, could we at least update the link to point to [the raw markdown](https://raw.githubusercontent.com/openjs-foundation/cross-project-council/master/NEW_PROJECT_APPLICATION.md) so that people can easily copy the list and paste it into the issue? Otherwise, it adds a step for people to find what they actually have to copy and could also lead to lists copied from the rendered HTML which isn't useful.

mcollina:
I don't see a big problem with the duplication, we are likely not updating this very often.

mhdawson:
My suggestion is to remove the checklist from the proposal doc instead and add a link to the template there.  That way when you create new issue you get the checklist and it is still discoverable through the proposal doc,

tobie:
The other option would be to clearly mark which of the two checklists is the canonical one.

MylesBorins:
I've made two changes.

1. Add onboarding checklist back to template
  - added comment about when it had been copied and where from which should remind people to update
2. I've moved the onboarding checklist from `./NEW_PROJECT_APPLICATION.md` to `./PROJECT_PROGRESSION.md` 

One thing to keep an eye out for... it seems like the copy from the application and the issue are already out of sync. Which version is canonical?

MylesBorins:
I plan to land this after today's meeting if there are no objections

PR #306, opened 2019-08-20
Add impact and consumers of the project to new project template

sendilkumarn:
Based on the [openjsf mail thread](https://lists.openjsf.org/g/new-projects/message/37).

This will be a good thing to add in the template and also will give both the project and our members to easily get the information when adding a new project. 

_p.s. cannot find anything similar after a quick search in the repository._


NEW_PROJECT_APPLICATION.md
```diff
@@ -13,12 +13,20 @@ A rough description of the project in less than 100 words.
...
+- What is the potential long term impact and adoption? 
+- How does it differ from other projects?
+
 ## List of all repos that are part of the project
 
 For each repository, please provide:
 
 - link to repo,
 - license information,
+- the IP belongs to,
```
dmethvin: "who owns the IP?"

NEW_PROJECT_APPLICATION.md
```diff
@@ -13,12 +13,20 @@ A rough description of the project in less than 100 words.
 Please refer to [the Cross Project Council's Charter](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md).
 Please keep your answer to less than 250 words.
 
+## Impact and consumers of the project
+- What is the ecosystem impact of the project? 
+- Who are the biggest consumers of the project? 
+_if they are not JavaScript space, then will it ever expand to JavaScript space_.
```
dmethvin: Need a bullet here? Also I'm wondering if we'd consider a project that wasn't using JavaScript.

NEW_PROJECT_APPLICATION.md
sendilkumarn: I thought of this but none of other things were a question in that list.

NEW_PROJECT_APPLICATION.md
sendilkumarn: >Need a bullet here? 

It is a continuation and not a separate question.

NEW_PROJECT_APPLICATION.md
ljharb:
```suggestion
- who owns the IP
```
this is still more correctly phrased, and isn’t in question form

sendilkumarn: Thanks @ljharb

proposals/stage-3/PROJECT_ONBOARDING/NEW_PROJECT_APPLICATION.md
```diff
@@ -13,12 +13,20 @@ A rough description of the project in less than 100 words.
 Please refer to [the Cross Project Council's Charter](https://github.com/openjs-foundation/cross-project-council/blob/master/CPC-CHARTER.md).
 Please keep your answer to less than 250 words.
 
+## Impact and consumers of the project
```
MylesBorins: This does not need to be modified (and arguably shouldn't). It is a historic document

sendilkumarn: oops, I thought it is duplicated. Will revert this.

eemeli:
I'm not sure that I'm happy with the term "consumer" here, as it's strongly (at least to me) implying a one-way relationship. I think "user" might be more fitting, especially as I consider my own role or that of other projects with respect to the existing OpenJS projects.

sendilkumarn:
@eemeli fair point. I agree. I will change this.

NEW_PROJECT_APPLICATION.md
```diff
@@ -13,12 +13,19 @@ A rough description of the project in less than 100 words.
...
+- What is the potential long term impact and adoption? 
+- How does it differ from other projects?
+
 ## List of all repos that are part of the project
 
 For each repository, please provide:
 
 - link to repo,
 - license information,
+- who owns the IP
```
mcollina: Isn't this part of the license?

sendilkumarn: I don't think so. But certain project's IP might be tied to their employer and it will be good to know this at the initial stage itself.

kborchers: I agree that asking this question at this point is important. I would suggest saying `Intellectual Property` as opposed to just `IP` to be clear.

NEW_PROJECT_APPLICATION.md
timmywil: > Also I'm wondering if we'd consider a project that wasn't using JavaScript.

This should probably be addressed. I'm not sure the answer here either.

kborchers: > It is a continuation and not a separate question.

Perhaps this should be added into the question to help clarify that instead of having it on a separate line? So something like `Who are the biggest consumers of the project and if they are not JavaScript space, then will it ever expand to JavaScript space?`

> Also I'm wondering if we'd consider a project that wasn't using JavaScript.

This would depend on what you mean by "using" JavaScript. IMO, anything with some sort of relationship with the JavaScript ecosystem would be a potential candidate. For example, libuv is already an OpenJSF project but is not solely a JavaScript-focused project and does not have JS in it's source (or at least not that I am aware of in the limited knowledge I have of that project's inner workings).

PR #309, opened 2019-08-21
doc: incubating projects must have a champion

MylesBorins:
We had originally had "identifying a champion" as a section in the
application but removed it to make the process more inclusive.

This change re-adds the champion as an entrance criteria for the
incubation phase. Without a champion in incubation it will be quite
a bit harder for projects to be successful. Further, if no member
of the CPC is willing to sponsor / champion a project that would
be a fairly clear sign that the project should not enter incubation.

### DIFF
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -20,6 +20,7 @@ New projects should express interest to join the OpenJS Foundation via an email 
 
 1. Initial email sent to new-projects@lists.openjsf.org with filled out [Project Application Template](./NEW_PROJECT_APPLICATION.md).
 1. Silent period. Internal only discussion.
+1. CPC champion is identified.
 1. Initial acceptance as [Incubating][] Project.
 1. Project goes through process of adhering to on-boarding checklist from [Project Application Template](./NEW_PROJECT_APPLICATION.md).
 1. Project presents to CPC for final acceptance via 2/3 supermajority vote and appropriate initial stage is determined.
@@ -149,7 +150,7 @@ Incubating projects are projects that are in the process of completing their on-
 
 #### Expectations
 
-It is expected that incubating project will make an active effort to work through the on-boarding process in a reasonable timeframe. They will be offered foundation support to help do so.
+It is expected that incubating project will make an active effort to work through the on-boarding process in a reasonable timeframe. They will be offered foundation support to help do so. All incubating projects will have a CPC champion.
 
 #### Acceptance Criteria
 


kborchers:
> Further, if no member of the CPC is willing to sponsor / champion a project that would be a fairly clear sign that the project should not enter incubation.

I'm not against having a champion be part of the entrance criteria but I don't fully agree with this assertion. What is the role of the champion? What sort of time investment is expected from a champion? Is it as simple as making a statement that the champion recommends this project or more of a mentor type role or something in between? Depending on the answers to these questions, it could limit the number of CPC members that are willing/able to champion a project through the incubation process. This could lead to rejecting a project that is a good fit but not "exciting" enough to get a CPC member to commit as a champion.

MylesBorins:
@kborchers I'm imagining that a champion is an individual taking on the responsibility of making sure that a project is successful. This would include navigating internal conversations, mentoring through the process, and potentially a number of other things that come up during the process.

We could loosen the requirements so that the champion doesn't need to be a CPC member... but someone who wants to be a champion can easily become a Regular Member by opening a PR.

I've added the agenda label so we can discuss in the next meeting.

jorydotcom:
I see the role of 'champion' as being akin to 'customer success' - the project ought to have someone they can count on to answer questions, help navigate process, and make sure things aren't slipping through cracks. I see it as a way to be welcoming and help a new project integrate with the community and meet new people. 

Right now it seems like these responsibilities largely fall in myself & @brianwarner's laps, and I'm fine if this role is by default one that's filled by staff/program people rather than CPC members. However that may become a challenge over time if budget doesn't support it.

tobie:
Worth noting the ASF [has a similar incubation process](http://incubator.apache.org/cookbook/#finding_a_champion_and_mentors) and has both the concept of champions (which help prepare application for the incubation stage) and mentors (up to 3 per project) whose role is to accompany the project up to graduation. The role and responsibilities of mentors are described [here](https://incubator.apache.org/guides/roles_and_responsibilities.html#mentor). Mentors must be members of the Incubator Project Management Committee which contains roughly [300 members](http://home.apache.org/phonebook.html?ctte=incubator).

PROJECT_PROGRESSION.md
```diff
@@ -149,7 +150,7 @@ Incubating projects are projects that are in the process of completing their on-
 
 #### Expectations
 
-It is expected that incubating project will make an active effort to work through the on-boarding process in a reasonable timeframe. They will be offered foundation support to help do so.
+It is expected that incubating project will make an active effort to work through the on-boarding process in a reasonable timeframe. They will be offered foundation support to help do so. All incubating projects will have a CPC champion.
```
tobie:
```suggestion
It is expected that incubating project will make an active effort to work through the on-boarding process in a reasonable timeframe. They will be offered foundation support to help do so. All incubating projects will have a CPC champion assigned by the CPC during the silent period.
```

mhdawson:
I think having a champion is important, Ideally, it should be somebody who can help the project through the process but I can see that with CPC members coming from existing projects that might be a challenge.

tobie:
BTW, the [application form](https://github.com/openjs-foundation/cross-project-council/blob/master/NEW_PROJECT_APPLICATION.md#project-champion) also mentions a Champion, which is described as "the main point of contact during the application process" which I assume is someone on the project side (and not on the CPC side). We'll want to fix that language accordingly.

MylesBorins:
@mhdawson can this be closed in lieu of #321?

mhdawson:
Yes, I'll close, please re-open if you don't think that is the right thing to do.

Issue #296, opened 2019-08-15
Specify the scope of CPC votes during application and incubation


tobie:
During the silent period, the CPC votes to accept the application of a project to join the OpenJSF. One the project is ready to leave incubation, the CPC votes again to allow the project to graduate. Additionally, the CPC can request that a project withdraw at any point of the incubation phase without having to provide justification for it.

As I understand it, the point of the silent period is to make sure there's a fit between a project and the foundation and to have room to reject an application if that's not the case, without having to publicly hurt people's feelings, the image of a project, or that of the corporations backing it.

Once this fit is established and the project's application has been accepted, it feels that the reasons to request withdrawal should be scoped and the request motivated. Requesting a withdrawal should be limited to projects breaking rules or having stalled on their path to graduation and not making concrete steps to get back on track.

Similarly, once a project is ready for graduation, the CPC's votes should be limited to decide whether the project meets graduation criteria. Other concerns, such as fit, should have been brought up during the initial vote.

Joining a foundation is a large investment for a project and the people behind it. It would be unfair to allow a project to invest important resources into this process only to have the CPC refuse graduation or request withdrawal on motives the project has no control over and which weren't specified in the graduation criteria or onboarding checklist.

mhdawson:
> Once this fit is established and the project's application has been accepted, it feels that the reasons to request withdrawal should be scoped and the request motivated. Requesting a withdrawal should be limited to projects breaking rules or having stalled on their path to graduation and not making concrete steps to get back on track.

I'm not quite sure I agree with this.  I agree the quiet period is to make sure the application does not proceed unless the CPC members think there is a good change of success.  Having said that, the CPC is a smaller number of people and additional information/feedback may come to light once the application is public and we want that public feedback to influence the decision...

My opinion is that the quiet period is an important part to avoiding the situation which nobody wants (application being refused after being submitted) but that it should not be a "done deal" after the submission with only "bad actions" on the part of the project being a valid reason for graduation.  If more information/further discussion means that it now understood that the project is not a good fit the CPC should make that hard decision and be trusted to do so.

tobie:
Thank you, @mhdawson, for your comment. That helps me better understand how folks on the CPC perceived the process and what their perspective is.

Could the CPC maybe define "good fit" in a way a project can self-assess before kick-starting the procedure? Right now, this seems really subjective and completely dependant on whoever sits on the CPC. Are their existing guidelines to frame that conversation?

[...]

PR #319, opened 2019-09-03
doc: expectations for dissent / objections

MylesBorins:
This adds language about the expectation for members to express
objections or dissent as early as possible.

Closes: https://github.com/openjs-foundation/cross-project-council/issues/296

### DIFF
--- a/MEMBER_EXPECTATIONS.md
+++ b/MEMBER_EXPECTATIONS.md
@@ -7,7 +7,8 @@ It is understood that members will have individual opinions and will express
 their individual opinions during discussions within the organization. All
 contributors are expected to be respectful of these different viewpoints,
 and to work collaboratively with one another in a manner that constructively
-elevates all contributors. If a member has an objection or dissent they should express it as early as possible to ensure there is ample time to discuss and reach consensus.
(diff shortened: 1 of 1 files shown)


MEMBER_EXPECTATIONS.md
```diff
@@ -7,7 +7,7 @@ It is understood that members will have individual opinions and will express
 their individual opinions during discussions within the organization. All
 contributors are expected to be respectful of these different viewpoints,
 and to work collaboratively with one another in a manner that constructively
-elevates all contributors.
+elevates all contributors. If a member has an objection or dissent they should express it as early as possible to ensure there is ample time to discuss and reach consensus.
```
mcollina: I would prefer to soften the language here, from “should” to “it is preferred” or “it is recommended”. 

I think we should allow objections to be brought up as late as possible, as people may travel/be off/whatever. Keeping it vague (asap) would allow some possible misunderstanding. Somebody might wonder “am I too late to object?”.

ljharb: If they’re traveling, then it’s as soon as possible whenever they’re back - this already seems very soft to me.

PR #341, opened 2019-10-03
Add nvm to at-large projects

brianwarner:
Adding nvm link to list of at-large projects.  Welcome, nvm!


MylesBorins:
I'm pretty sure this should be under the `Incubation` umbrella not `at-large`

My impression is that this is not "officially" in the foundation yet... but that is a bit pedantic. That being said we should likely be careful about that language before IP transfer and acceptance vote has gone through.

Pedantry aside this is exciting! Congrats nvm!

brianwarner:
@MylesBorins you are very right, this was my error.  My apologies...  I should have added an "Incubating" category and welcomed nvm as the first project under that.  I'm editing the title of this PR, and will push a new patch in a moment.

PR #347, opened 2019-10-10
Fix https -> http link for nvm

kborchers:


kborchers:
Requesting fast track approval

kborchers:
With 2 fast track and 2 review approvals I will land this PR. Thanks!

ChALkeR:
A better solution would be for nvm.sh to enable https support...

brianwarner:
Assuming everything proceeds through the process, SSL forwarding is something which the Foundation can provide for managed domains.

In the meantime @ljharb, if it would be helpful I can offer some pointers in setting up Let's Encrypt for a forwarding-only domain.  Catch me on Slack if I can help.

PR #384, opened 2019-10-29
[meta] remove stray quote in CoC

ljharb:


eemeli:
This should be fast-tracked.

PR #385, opened 2019-10-29
Add Christian Bromann (@christian-bromann) as regular member

christian-bromann:
Adding myself as regular member to CPC as well as WebdriverIO to the list of growth projects.

😊

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ According to the [CPC Charter](https://github.com/openjs-foundation/cross-projec
 - Myles Borins (@MylesBorins)
 - Sendil Kumar (@sendilkumarn)
 - Tierney Cyren (@bnb)
+- Christian Bromann (@christian-bromann)
 
 ### Observers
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -88,6 +88,7 @@ According to the [CPC Charter](https://github.com/openjs-foundation/cross-projec
 - **Mocha**: Christopher Hiller (@boneskull)
 - **Node-RED**: Nick O'Leary (@knolleary)
 - **webhint**: Antón Molleda (@molant)
+- **WebdriverIO**: Christian Bromann (@christian-bromann)
```
eemeli: This is not a list of projects, but a list of voting members of the CPC that represent growth & at-large project (see the preceding paragraph).

eemeli: To clarify, these representatives were chosen back in May. Next spring we'll organise another selection, and should only end up with two representatives in this category.

PR #389, opened 2019-11-01
Update contact email for new project proposals

brianwarner:
I would like to propose a modification to the process used when proposing new projects.  I believe this is a low impact change, and would like to put it up for discussion via this PR.

This commit changes the email address to which new projects proposals are sent.

Currently, new project proposals are sent directly to the mailing list where
they are also discussed.  By having the proposal first come to a person, it
provides a first-round filter to ensure all information in the application is
complete, and also ensures a clean separation between the project applicant and
the committee which is discussing the application.  The OpenJS Foundation
program manager (currently me) would then be responsible for forwarding
applications on to the (private) discussion list.

### DIFF
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -14,7 +14,7 @@ This governance policy sets forth the proposal process for projects to be accept
 
 ### Project Proposal Requirements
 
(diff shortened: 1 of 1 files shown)


knolleary:
+1 on this - every time I reply on the new-project list I have to triple check I'm not replying to an email with the original proposer on copy. This would provide a fire-break to ensure no-one accidentally does this (other than Brian...).

We would need to make sure vacation periods are covered so new applications aren't sat in an inbox waiting for someone to come back from vacation - but I assume that's already well covered for other operational matters.

brianwarner:
My thoughts exactly, @knolleary.

And yes, per your comment about vacations and such - one of the primary reasons for using operations@openjsf.org for just about everything is that I can easily redirect emails to other people if I'm unavailable, without anybody having to change their processes or address books.

joesepi:
Expectations discussed in today's CPC meeting (#395):
- new email address for new applicants to send their applications is operations@openjsf.org
- person managing that email alias will:
  1) confirm application completeness
  2) acknowledge application receipt to applicant
  3) identify project manager
  4) forward application to private discussion email list

PR #407, opened 2019-11-14
Changes to CoC Handling

nainar:
Changes made: 
- s/reportee/reporter or target
- Add a link to CoC

### DIFF
--- a/proposals/stage-2/CODE_OF_CONDUCT/HANDLING-REPORTS.md
+++ b/proposals/stage-2/CODE_OF_CONDUCT/HANDLING-REPORTS.md
@@ -1,7 +1,7 @@
 # Handling reports and escalation
 This document describes the proposed process for handling reports and escalation, along with relevant roles in the OpenJS Foundation.
 
-This process covers two types of reports based on ./FOUNDATION_CODE_OF_CONDUCT_REQUIREMENTS.md:
+This process covers two types of reports based on the [Foundation's Code of Conduct Requirements](https://github.com/openjs-foundation/cross-project-council/blob/master/proposals/stage-2/CODE_OF_CONDUCT/FOUNDATION_CODE_OF_CONDUCT_REQUIREMENTS.md):
 
 * Reports for spaces managed by the Cross Project Council (CPC) which come in through `report@opensjsf.org`
 * Escalations which come in through `coc-escalation@openjsf.org`.   
@@ -10,10 +10,15 @@ Generally, reports to coc-escalation should refer to a report that was already s
 
 * report@openjsf.org or
 * one of the project's reporting emails 
-* explanation of why the reportee does not feel comfortable using the previous reporting emails for the initial report
+* explanation of why the reporter does not feel comfortable using the previous reporting emails for the initial report
 
-If a report to coc-esclation does not include one of those, the reportee will be asked to to provide one of those three as the 
+If a report to coc-esclation does not include one of those, the reporter will be asked to to provide one of those three as the 
 reason for reporting to coc-escalation.
+
+## Code of Conduct
+
+The OpenJS Foundation and its member projects use the Contributor Covenant v1.4.1 as its Code of Conduct. Refer to the [Code of Conduct](https://github.com/openjs-foundation/cross-project-council/blob/master/CODE_OF_CONDUCT.md) for the full text and reporting and esclation procedures. 
+
 
 ## Confidentiality and record-keeping
 Personal information is confidential. All reports should be recorded, together with the discussion of it. The following private repos will be used to record and discuss reports.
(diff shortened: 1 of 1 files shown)


proposals/stage-2/CODE_OF_CONDUCT/HANDLING-REPORTS.md
```diff
@@ -29,52 +34,40 @@ Any member of the CPC or Code of Conduct Panel(CoCP) who is involved in the repo
...
-        step 5 be repeated as well if the proposed action is modified.
-     6. Final resolution: the resolution agreed by the CoCP members is implemented and the outcome reported back to the reportee by
-        the main contact.
+        project leadership). If the leadership representatives expresses concerns then the contact person will bring these back to the CoCP members for further discussion.  This cycle can continue until the CoCP members reach consensus that the leadership representatives concerns have been adequately addressed.
+     6. Resolution shared: The resolution is shared with the reporter or target by the contact person. If the reporter or target expresses concerns then the contact person will bring these back to the CPC members for further discussion. This cycle can continue until the CPC members reach consensus that the reporter or target's concerns have been adquately addressed. If no feedback is received within 7 days the resolution is considered as accepted. 
+     6. Final resolution: the resolution agreed by the CoCP members is implemented and the outcome reported back to the reporter or target bythe main contact.
```
molant:
```suggestion
     7. Final resolution: the resolution agreed by the CoCP members is implemented and the outcome reported back to the reporter or target bythe main contact.
```

Or we could just number everything `1.` and let markdown renderers take care of the numbering.

proposals/stage-2/CODE_OF_CONDUCT/HANDLING-REPORTS.md
```diff
@@ -10,11 +10,16 @@ Generally, reports to coc-escalation should refer to a report that was already s
...
-* explanation of why the reportee does not feel comfortable using the previous reporting emails for the initial report
+* explanation of why the reporter does not feel comfortable using the previous reporting emails for the initial report
 
-If a report to coc-esclation does not include one of those, the reportee will be asked to to provide one of those three as the 
+If a report to coc-esclation does not include one of those, the reporter will be asked to to provide one of those three as the 
 reason for reporting to coc-escalation.
 
+## Code of Conduct
+
+The OpenJS Foundation and its member projects use the Contributor Covenant v1.4.1 as its Code of Conduct. Refer to the [Code of Conduct](https://github.com/openjs-foundation/cross-project-council/blob/master/CODE_OF_CONDUCT.md) for the full text and reporting and esclation procedures.
```
tobie: I think that the version number is just 1.4.

```suggestion
The OpenJS Foundation and its member projects use the Contributor Covenant v1.4 as its Code of Conduct. Refer to the [Code of Conduct](https://github.com/openjs-foundation/cross-project-council/blob/master/CODE_OF_CONDUCT.md) for the full text and reporting and esclation procedures. 
```

tobie: Note I think this is also incorrect in the CoC itself (and could be fixed as part of this PR).

bnb: I believe it's actually 1.4.1, the versioning on the website slug / repo just isn't particularly clear. See: https://www.contributor-covenant.org/#contributor-covenant-v1-4-1

Additionally, 2.0 seems to have shipped: https://www.contributor-covenant.org/version/2/0/code_of_conduct

ljharb:
It'd be ideal to use `1.` for all numbered list items; then we never have to keep track of the numbering.

nainar:
Changed the numbering to be all `1.`

proposals/stage-2/CODE_OF_CONDUCT/HANDLING-REPORTS.md
eemeli: This might be a bit pedantic, but our `CODE_OF_CONDUCT.md` document does not in fact contain "the full text" of our code of contact; instead it links to contributor-covenant.org for the full text.

Could we add the English-language text to our own CoC document, and link to the source for the translations?

ljharb: Why would we want the english text to be saddled with a redundant source of truth, while the translations remain properly single-source-of-truth-ed?

eemeli: First of all, for ease of use and to match the expectations of the covenant text itself; it includes this near the end:
>This Code of Conduct is adapted from the Contributor Covenant, version 1.4, available at https://www.contributor-covenant.org/version/1/4/code-of-conduct.html

Second, because by referring to and trusting only an external source, we run a risk that said external source will change the text. This might sound like a theoretical risk, but the first "version 1.4.1" of the covenant [was published](https://github.com/ContributorCovenant/contributor_covenant/commit/5738a899ea6374469396536a0a7fabe5a96c7fb7) in April 2017, and since then it's been modified by [at least these commits](https://github.com/ContributorCovenant/contributor_covenant/commits/release/content/version/1/4/code-of-conduct.md). Those are not all cosmetic, as categories of characteristics have been added to the initial pledge and the scope has been amended, all without changing the version number.

Now that version 2.0 of the covenant has been published, it's of course less likely that 1.4 will be further amended, but it's not impossible either. Having the actual text under our control would allow us to be clear about what it actually says.

ljharb: Perhaps then we should link to both the web version, and the git sha?

If we're going to inline the text, then why link elsewhere at all? Your concerns apply to the translated text just the same as to the english text.

nainar: I think the concern with having the CoC inlined was that it could get out of date and we would have to update that every time we made changes. 

I can make changes to the text inline to say "Go here to find a link to the CoC text"

eemeli: I would actually count it as a good thing for us to need to be explicit about updating our CoC when it changes. And we'd need to do that even when linking to the text, as its version number would hopefully change on updates.

But sure, this is a separate issue if the inline text here is updated to reflect the current situation.

eemeli: Added #413.

mhdawson:
@eemeli are you still objecting to this landing.  Your comment is resolved but you still have a blocking review open.

mhdawson:
Pushed commit to address last comment. We agreed it could land with that being updated.

Issue #382, opened 2019-10-29
CPC LICENSE.md references Node.js Contributors


joesepi:
https://github.com/openjs-foundation/cross-project-council/blob/master/LICENSE.md

```
MIT License (MIT)

Copyright (c) Node.js contributors. All rights reserved.

Permission is hereby granted, free...
```

PR #414, opened 2019-11-26
Add Foundation-wide copyright guidance

brianwarner:
This PR adds Foundation-wide copyright guidance on how to add copyright
notices to source code headers.  This document is based on the guidance document
from CNCF, available here:
https://github.com/cncf/foundation/blob/master/copyright-notices.md

To be absolutely clear, do *NOT* modify an existing copyright line unless you put it
there, or have explicit permission from the one who did.  This guidance only
applies to copyright notices which are *APPENDED* to the existing list of
notices.

In addition, the OpenJS Foundation Board of Directors recently approved
standardized website footers for projects to use.  These include the copyright
notice, as well as trademark notices and links to Foundation-wide policies.  One
notice is for projects related to Node.js, and includes a line referencing the
trademark license.  The other is for projects which are unrelated to Node.js,
and omits the line.

If you have any questions, please reach out to legal-questions@openjsf.org

Signed-off-by: Brian Warner <brian@bdwarner.com>

PR #434, opened 2019-12-20
doc: update license copyright to OpenJS foundation

MylesBorins:
This is a relic of the repo being created in the Node.js foundation org

My assumption is this should be non controversial but want a gut check.

Closes: https://github.com/openjs-foundation/cross-project-council/issues/382

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 MIT License (MIT)
 
-Copyright (c) Node.js contributors. All rights reserved.
+Copyright (c) OpenJS Foundation and contributors. All rights reserved.
 
 Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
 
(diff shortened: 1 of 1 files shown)


MylesBorins:
We should get feedback from foundation folks before fast tracking this

molant:
In this conversation with @brianwarner we talked about this same issue for webhint. it's OK to replace with openjs:

https://github.com/openjs-foundation/cross-project-council/pull/414#discussion_r353510795

brianwarner:
In this particular instance, I would recommend adding "OpenJS Foundation" in addition to "Node.js contributors."

Here's my rationale, since @molant referenced the other thread and I'd like to avoid creating any confusion in this response.  The general guidance is that the text "JS Foundation" and "Node.js Foundation" can be updated to "OpenJS Foundation", since it is documenting a fact that the two orgs are merged and renamed.

Here, the line just says "Node.js contributors".  Since that could mean either project or Foundation, I would recommend the almost-always-safe path of appending rather than replacing.

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License (MIT)
 
-Copyright (c) Node.js contributors. All rights reserved.
+Copyright (c) OpenJS Foundation and contributors. All rights reserved.
```
brianwarner:
```suggestion
Copyright (c) Node.js contributors. All rights reserved.
Copyright (c) OpenJS Foundation and contributors. All rights reserved.
```

summerrobert7281:
Wasn't it spelled differently though like fastrak or something..??

PR #440, opened 2020-01-07
doc: document end date for CPC Director term

mhdawson:


mhdawson:
Has met the criteria for landing.

Issue #377, opened 2019-10-22
Clarify Travel Fund status & process


eemeli:
The travel fund proposal got advanced to Stage 3 in #312, and is therefore nominally "complete". However, its documents are only included under `proposals/stage-3/TRAVEL_FUND`, they contain a number of "TODO" mentions, and at least the 2019 record is missing some updates that have been made to the [nodejs/admin version](https://github.com/nodejs/admin/blob/master/TravelFunds/2019.md).

As an interested party to applying for the travel fund for the Montreal conference & summit, I actually have no idea about how I should proceed with my application. Should I apply here, or in the nodejs/admin repo? When is the process moving to this repo? How will any open requests be handled?

Jonahss:
Open requests can be migrated. So apply through the old repo until #312 is fully completed.

The process itself (using expensify, email addresses) are all the same.

Jonahss:
I plan on setting up the final PR for the proposal in the next 7 days. So if you want to wait for that, you can.

christian-bromann:
Given that we were short in time at the CPC meeting, let me phrase my question in here:

- There is a section at the end on approved travel funds of previous year. Do we have already approval for travel fund this year? If so, by whom? Is it the OpenJS Foundation this time?
- Shall I make changes to the "complete" stage 3 proposal or should we move this back to an earlier stage?

christian-bromann:
@eemeli please review final PR with suggested changes: https://github.com/openjs-foundation/cross-project-council/pull/441. Cheers!

mhdawson:
Landed #441 so closing.

PR #441, opened 2020-01-07
Finalise travel fund process

christian-bromann:
This PR finalises the in stage 3 located proposal for travel funding. There are still some outstanding questions in https://github.com/openjs-foundation/cross-project-council/issues/377 which need to be answered first. Also I want get @brianwarner 's input as we recently had an event where people request funds.


christian-bromann:
@mcollina yes, I raised the question in the [related issue thread](https://github.com/openjs-foundation/cross-project-council/issues/377#issuecomment-571754770) and hope to get this answered.

mhdawson:
@mcollina I suggest you talk to @MylesBorins as his take was that the existing budget was enough and we did not need to pre-allocate more until it was needed (@MylesBorins please correct me if I'm misremembering on this front).  I had raised the issue at the board level as to whether we should increase the budget but the discussion ended up trending towards that the current budget was likely enough and we could ask for more if/when it ran out.

MylesBorins:
I did some rough addition of the 19' usage of the fund and afaict we used just over 30k of the 60k budget. The reason to not increase the budget was because we explicitly did not need to. It is worth noting that the 19 budget was utilized by all projects, not just Node.js.

If we run out of budget the board can approve an increase to the budget. My understanding is that this fund covering more than just the Node.js project would be a very good reason to approve of raising the budget. To my knowledge there is no reason why the board would not approve an increase if we needed it.

I'd say the reason to move it is not solely an increase of budget. We are part of a larger foundation and not the only individuals utilizing it. The Node.js project benefits from other sharing the task of administering the budget, it makes the process more open, and we maintain our ability to participate in the budget process via CPC (which is open to anyone from any project).

I really see 0 reason to maintain the budget within the project and if you are going to -1 this transfer would very much appreciate a clearer outline of why you think this is benefiting the project. Node.js maintaining it's own budget would, at least to me, give the impression of the project not trusting or wanting to work with other foundation projects... which is not the message I would like to see our project sending to the foundation.

mcollina:
I'm sorry for the short note. Here is a longer explanation.

The '19 budget has been marginally used by other projects, and with all the new projects joining I expect a significant increase in usage. I see this as a key function of this foundation, and there should be enough money to cover for a large Summit and more.

We had a lot of discussion in the summer of '18 about the merger, and we got plenty of reassurances that OpenJS will keep supporting Node.js at the same level. From my point of view this looks like a reduction considering that there should be more funding available post-merger.

I'm citing @trott from https://github.com/openjs-foundation/cross-project-council/issues/172#issuecomment-490523657:

> I'd recommend being very careful in the approach taken here. The travel fund was set up by the Node.js Foundation board to promote Node.js and to be administered by the Node.js TSC and CommComm. Basically negating that and making it an OpenJS thing would have the appearance of confirming the fears of some on the TSC that the foundation merger would dilute Node.js resources.

This has been a valid concern that was brought on several times by the TSC during the merger process. I feel it's my responsibility to note this is not following what has been discussed by leaders in the Node.js community on multiple occasions over the past year and a half.

I'm happy to bring this point to the TSC and CommComm for confirmation.

jasnell:
I have to agree with @mcollina, this feels like a step backward. I'm currently -1 on it also.

timmywil:
I noticed a couple things while observing this conversation.

1. Maybe I'm reading wrong, but @mcollina seems to be thinking of this more like merging bank accounts, where the final balance should be the sum of all accounts, but a budget doesn't work like that. The budget *can* stay exactly the same as long as it covers everyone post-merger. I get this from the "this looks like a reduction" comment. It is not a reduction. It is staying the same with the expectation that is already enough even with more people using it.
1. It's difficult to know ahead-of-time how much a fund will get used. I think @MylesBorins and @mcollina agree that some increase in usage is expected (exactly how much is unpredictable). The disagreement is about whether the current budget will cover that increase in usage without changing. If @mcollina doesn't think it will, that's a valid concern. Unfortunately, the only way to know for sure is to try it and see. Then, if all of it is going to be used up, adjust from there. As long as all parties agree that we can increase the budget later, I don't see the problem.

My main point is this: raising the budget only for appearances sake is not a good reason. Raise the budget if you actually think it needs to be raised.

joesepi:
I was under the same impression that @mhdawson and @MylesBorins are expressing here -- that we had agreed to keep the budget the same since we only used roughly half of the 2019 budget. In my experience you don't ask for more budget if you didn't use it all last year.

**However**, when we were previously discussing the travel fund, we didn't know we would bring on two large projects such as Electron and AMP, both of which have their own collaborators summits.

Now we are in this position where we are looking at the convergence of the following:
- finalizing travel process and 2020 travel fund budget
- have 2 large projects having joined the foundation
- have an upcoming budget discussion
- have a half unused 2019 budget

Here is what I think is reasonable in this situation considering that in 2019, roughly $30k was used, with the majority of that being Node.js but not completely:
Allocate $25k each for the 3 large projects that currently have collaborator's summits (Node.js, Electron, AMP) plus another $15k for all other collaborator/contributor travel requests.

This would equal $90k, which is 50% more than 2019 budget ($60k) but does not seem to be an unreasonable ask. This doesn't water down Node.js use of the travel fund and also makes this great resource available to our other projects.

dmethvin:
This PR says:

> Those requesting funds must indicate that they do not have funding available from another
  source, such as an employer or the event itself that might cover costs for presenters.

Electron has a [$25,000 annual budget](https://opencollective.com/electron#section-budget) in OpenCollective. AMP has had [$117,000 contributed](https://opencollective.com/AMP). I am not sure why projects with such large self-managed sources of funds would be drawing from the communal travel fund.

MEMBER_TRAVEL_FUND.md
```diff
@@ -0,0 +1,178 @@
...
+## Process
+
+Requests must be made prior to the start of travel. Requests submitted once travel is
+complete will not be approved.
+
+It is strongly recommended that you submit your request far enough in advance to get
+approval before making relevant expenditures. Review of requests will take at least 72
+hours and may take considerably longer. Note, with limited funds available, approval
+is not guaranteed. If a request is rejected, reimbursement will not be issued. The
+Foundation cannot pre-pay for travel expenditures.
```
jasnell: What I would prefer here is that requests made in regard to specific project events are still directed to the steering bodies for each project, rather than to the CPC. The reason is those steering committees will have the most context for determining approval of a request.

For example: If I am attending a Node.js Diagnostic Working Group face-to-face and want to request travel funds, that request should go to the Node.js TSC/CommComm for approval as it is now. If I am attending a Foundation event (such as an OpenJSF Collaborator's Summit) then the request would go to the CPC.

For those who use corporate expense reimbursement type systems, it's no different really than having different approvers for expenses that have different tags.

In advance of a project specific event, the project should submit a request to the CPC for an earmarked allocation for that event. For instance, "Hey, we have a diagnostic working group face to face coming up, we expect about 20 people to attend, please earmark about $5000 for the event to cover travel requests".

jasnell:
If I understand @mcollina correctly, I think the budget concern is more about moving away from a bucket dedicated specifically to Node.js to one that is shared among many projects. The current allocation of $60k is currently more than sufficient if we're currently only spending $30k of it but not so if seven other projects pull $10k each out of the pot for their own events.

A possible way around that, at least in terms of making things more predictable/transparent is to ask that projects that have project-specific events coming up request an earmarked allocation in advance of the event, and in advance of individuals requesting reimbursement. (as I mentioned in my other comment here: https://github.com/openjs-foundation/cross-project-council/pull/441#discussion_r366648789).

MEMBER_TRAVEL_FUND.md
dmethvin: That makes sense. As I mentioned [above](https://github.com/openjs-foundation/cross-project-council/pull/441#issuecomment-574440617), many of these projects are independently collecting and managing their own funds. They should be tapping into those funds first and not using this travel fund. By contacting their own project committees they can determine if project funds are available before asking for foundation funds.

mhdawson: @jasnell this was discussed in the CPC meeting today. One suggestion is that we could land the current PR as is so that we can have the CPC approve travel requests that come in (versus the current state of confusion of how to get approval).  We could then work on an extension that incorporates what you suggested in terms of projects asking for allocations and then managing the approvals themselves.  The projects would then be able to make that ask and decide/implement the approval process in their own repos.

mhdawson:
@openjs-foundation/cpc I think we should have some additional CPC member approvals for this to land.

mhdawson:
As mentioned in the public section f the Board meeting today:

* Budget is confirmed at 60K. This is 20-30k more than was spent last year so we believe there will be enough to cover an increase in travel requests due to other projects using the fund
* The board issued this statement:
  * The OpenJS Foundation is committed to the travel fund and has allocated an amount based on
    past usage with what we believe is enough headroom to cover all travel in 2020. If it turns
    out the program is more successful than budgeted for, we are committed to provide an
    appropriate budget to cover the needs of all our projects.

I think we are good to go on the funding side.

mhdawson:
@jasnell I think your thumbs up on https://github.com/openjs-foundation/cross-project-council/pull/441#discussion_r369207294 is an indication that you are now ok with this landing as a first step.  Is that correct? Want to be sure as if so it would mean there are no longer any objections and we can proceed with landing it.

jasnell:
Yeah, I do want to see the additional changes explored tho

mhdawson:
@jasnell once this lands I'll take a cut at a PR to add the wording to allow projects to request funds as suggested and managed the approvals for those funds.

MEMBER_TRAVEL_FUND.md
```diff
@@ -0,0 +1,178 @@
...
+  * Include as many fields as possible:
+    * The name of the event you plan to attend.
+    * The location, dates of the event and where you will be traveling from.
+    * The presentation you intend to give, if applicable.
+    * The size of the stipend you wish to receive.
+      * In the pull request description, explain how you calculated your
+        stipend. For example, include estimates for airfare, hotel, other
+        transportation, and so on. The total amount should be your stipend
+        request.
+      * In the PR description, mention @openjs-foundation/cpc.
```
eemeli: Just as a note, I don't think this actually pings CPC members unless the requester is a member of the org?

MEMBER_TRAVEL_FUND.md
```diff
@@ -0,0 +1,178 @@
...
+    * Once the final amount spent is known, update the table again with
+      that information.
```
eemeli: In what currency units should this field be expressed, both before and after actually spending the funds? For instance, should a EU-based visitor to a Canadian event use EUR, CAD or USD here?

ljharb: I’d assume in whatever currency the amount is paid.

eemeli:
I believe this closes #377 when merged.

MEMBER_TRAVEL_FUND.md
mhdawson: From the past files it seems to have been either US or EU so I think @ljharb is correct.

mhdawson:
Landing as we have a number of approvals, no remaining objections and the waiting time for Governance docs has elapsed.  

There will be follow on PRs to address @jasnell suggestion and possibly the addition of some additional info

PR #463, opened 2020-02-12
doc: add new reg member to email list

joesepi:
This is a governance change and needs to be open for at least 14 days OR consensus is reached in a meeting with quorum of voting members.


joesepi:
If we decide to start sending new members a welcome email, we may want to update this PR reflecting that as well. Since this is a governance change, will be more efficient to bulk update these little changes. :) 

https://github.com/openjs-foundation/cross-project-council/issues/464

mhdawson:
@joesepi, while you say its a governance change, we've already been doing it right?  Just want to know if this is proposing a change in practice or documenting what we'd decided to do already.

joesepi:
True, @mhdawson -- we already add new reg members to email list and I am looking to merely document that. In a comment above, I mention a welcome email which would be an additive change, but that doesn't need to go along with this PR per se.

PR #483, opened 2020-03-05
chore: replace Dan Graham with Isaac Murchie as Appium representative

imurchie:
Dan Graham does not have time to commit to the Cross Project Council, so Isaac Murchie (@imurchie) will step in.

Cc: @dpgraham

PR #495, opened 2020-03-17
Amendment to the by laws to ensure a sign-off between Voting CPC members

christian-bromann:
This patch adds a section to the by laws to ensure that Voting CPC members connected to an impact or non impact project are being signed off by the current active CPC member of that project.

See also discussions in https://github.com/openjs-foundation/cross-project-council/pull/483#issuecomment-595432313

### DIFF
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -109,9 +109,11 @@ one-fourth of the Voting CPC membership to be affiliated with the same employer.
 Such a situation must be immediately remedied by the resignation or
 removal of one or more voting CPC members.
 
-If a Voting CPC member connected to an impact or non impact project is being
-replaced by someone from the same project, a sign off from the current active
-CPC member is required with an approval.
+If a voting CPC member steps down before the end of their one year term or they
+are removed from the CPC by a CPC motion they may be replaced using the same
+process by which they were elected. In this case the member will serve out the
+remainder of the term and must be re-confirmed at the end of the original one
+year term as if they had served for the full year.
 
 The public portion of the CPC discussions and meetings will be open to all
 observers and members.
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -109,6 +109,10 @@ one-fourth of the Voting CPC membership to be affiliated with the same employer.
 Such a situation must be immediately remedied by the resignation or
 removal of one or more voting CPC members.
 
+If a Voting CPC member connected to an impact or non impact project is being
+replaced by someone from the same project, a sign off from the current active
+CPC member is required with an approval of the new member.
+
 The public portion of the CPC discussions and meetings will be open to all
 observers and members.
 
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -111,7 +111,7 @@ removal of one or more voting CPC members.
 
 If a Voting CPC member connected to an impact or non impact project is being
 replaced by someone from the same project, a sign off from the current active
-CPC member is required with an approval of the new member.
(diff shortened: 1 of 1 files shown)


dmethvin:
A thought: Do we need a contingency for the case where the person being replaced can't be contacted?

christian-bromann:
@dmethvin makes absolutely sense. I would give this up for discussion.

In my opinion for a CPC voting member connected to an impact project, an approval could also be made by the other CPC member of that project.

In case the project only has one representative which can't be contacted:
- the replacement can be auto approved after a specific amount time (e.g. 30 days) or
- the request for replacement has to be denied and that person needs to wait until it can be nominated for that project like a regular member

CPC-CHARTER.md
```diff
@@ -109,6 +109,10 @@ one-fourth of the Voting CPC membership to be affiliated with the same employer.
 Such a situation must be immediately remedied by the resignation or
 removal of one or more voting CPC members.
 
+If a Voting CPC member connected to an impact or non impact project is being
+replaced by someone from the same project, a sign off from the current active
+CPC member is required with an approval.
+
```
mhdawson: I'm thinking maybe something like:

If a voting CPC member steps down before the end of their one year term they may be replaced using the same process by which they were elected. In this case the member will serve out the remainder of the term and must be re-confirmed at the end of the original one year term as if they had served for the full year.  

I think this is a bit more general, allows for a replacement and captures how long they will server for.

mhdawson: In terms of needing approval from the original member. I think once they have officially stepped down it's up to the project and we don't necessarily need their approval.

mhdawson: Edited my suggestion to also cover the case where they are removed by the CPC. This would hand the case where they can no longer be contacted.

mhdawson:
Going to mark changes requested for now.

CPC-CHARTER.md
dmethvin: @mhdawson I think your suggestion is great. I had already approved, I guess I could have withdrawn but I wouldn't have come up with good wording!

MylesBorins:
Haven't had a chance to review yet but high level FYI, if we want to change the charter this will require board approval. Do you think we could document this case in the governance instead (which doesn't require board approval)

kborchers:
I would prefer to see wording closer to @mhdawson's suggestion.

joesepi:
I'm also in favor of @mhdawson's language.

CPC-CHARTER.md
christian-bromann: Updated wording.

mhdawson:
@openjs-foundation/cpc, this is ready to go and needs approval by at least 1/2 of the voting members.

dmethvin:
I approved earlier, but reiterating my support for the latest version.

bnb:
per a reminder I set for myself during the last CPC meeting: can this be merged now? Seems like we have pretty strong consensus.

mhdawson:
@bnb as a change to the charter we need

1) 1/2 of the voting members to sign off (I think you were going check/chase that) Would have to check if 8 is 1/2 or not

2) Board approval. Once we have 1) I'll get the board vote started.

joesepi:
@mhdawson is sending a note to board for approval.

mhdawson:
@joesepi and @bnb  has confirmed that we have more than 1/2 of the voting members.

Michael has action to start board vote.

mhdawson:
Vote is in progress.

mhdawson:
Confirmed with Brian today that Board vote has approved. Will land

PR #508, opened 2020-03-31
Specify non-impact voting CPC member election timeframe

eemeli:
As discussed in today's CPC meeting, the governance doc specifies the nomination period when electing non-impact projects' voting CPC members, but does not specify any other related timeframes.

This year, we're not going to start voting until after the next term starts tomorrow. It would be nice if we could lock down this timeframe to more specifics to help us make sure that the current situation is not repeated in future years.

This PR suggests an initial one week period for receiving nomination issues, and then a voting period of two weeks.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -113,6 +113,8 @@ These members will be elected for a term of 1 year as follows:
 
 * Each voter will be able to vote for as many or as few candidates as they wish.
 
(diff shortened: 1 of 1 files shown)


joesepi:
This PR reaches the threshold needed to approve as a governance change.

https://github.com/openjs-foundation/cross-project-council/blob/master/GOVERNANCE.md#merging-prs-into-this-repository

Issue #500, opened 2020-03-23
OpenJS Foundation Cross Project Council Meeting 2020-03-24


mhdawson:
## Time

**UTC Tue 24-Mar-2020 16:00 (04:00 PM)**:

| Timezone      | Date/Time             |
|---------------|-----------------------|
| US / Pacific  | Tue 24-Mar-2020 09:00 (09:00 AM) |
| US / Mountain | Tue 24-Mar-2020 10:00 (10:00 AM) |
| US / Central  | Tue 24-Mar-2020 11:00 (11:00 AM) |
| US / Eastern  | Tue 24-Mar-2020 12:00 (12:00 PM) |
| London        | Tue 24-Mar-2020 16:00 (04:00 PM) |
| Amsterdam     | Tue 24-Mar-2020 17:00 (05:00 PM) |
| Moscow        | Tue 24-Mar-2020 19:00 (07:00 PM) |
| Chennai       | Tue 24-Mar-2020 21:30 (09:30 PM) |
| Hangzhou      | Wed 25-Mar-2020 00:00 (12:00 AM) |
| Tokyo         | Wed 25-Mar-2020 01:00 (01:00 AM) |
| Sydney        | Wed 25-Mar-2020 03:00 (03:00 AM) |

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Cross%20Project%20Council+Meeting+2020-03-24&iso=20200324T16
* or http://www.wolframalpha.com/input/?i=04PM+UTC%2C+Mar+24%2C+2020+in+local+time

## Links

* Minutes Google Doc: <https://docs.google.com/document/d/1m2gs5Vv5xs55kB26anj66TD7ZdsQueYRnB2pbCd8JXg/edit>

## Agenda

Extracted from **cross-project-council-agenda** labelled issues and pull requests from the **openjs-foundation org** prior to the meeting.

### openjs-foundation/summit

* Travel Fund Questions for Project Community Members [#249](https://github.com/openjs-foundation/summit/issues/249)

### openjs-foundation/cross-project-council

* \[Action Requested\] Nominations Sought for 2020-2021 Voting CPC Membership  [#488](https://github.com/openjs-foundation/cross-project-council/issues/488)
* webhint: project charter review [#482](https://github.com/openjs-foundation/cross-project-council/issues/482)
[...]

Issue #493, opened 2020-03-16
OpenJS Foundation Cross Project Council Meeting 2020-03-17


mhdawson:
## Time

**UTC Tue 17-Mar-2020 18:00 (06:00 PM)**:

| Timezone      | Date/Time             |
|---------------|-----------------------|
| US / Pacific  | Tue 17-Mar-2020 11:00 (11:00 AM) |
| US / Mountain | Tue 17-Mar-2020 12:00 (12:00 PM) |
| US / Central  | Tue 17-Mar-2020 13:00 (01:00 PM) |
| US / Eastern  | Tue 17-Mar-2020 14:00 (02:00 PM) |
| London        | Tue 17-Mar-2020 18:00 (06:00 PM) |
| Amsterdam     | Tue 17-Mar-2020 19:00 (07:00 PM) |
| Moscow        | Tue 17-Mar-2020 21:00 (09:00 PM) |
| Chennai       | Tue 17-Mar-2020 23:30 (11:30 PM) |
| Hangzhou      | Wed 18-Mar-2020 02:00 (02:00 AM) |
| Tokyo         | Wed 18-Mar-2020 03:00 (03:00 AM) |
| Sydney        | Wed 18-Mar-2020 05:00 (05:00 AM) |

Or in your local time:
* http://www.timeanddate.com/worldclock/fixedtime.html?msg=Node.js+Foundation+Cross%20Project%20Council+Meeting+2020-03-17&iso=20200317T18
* or http://www.wolframalpha.com/input/?i=06PM+UTC%2C+Mar+17%2C+2020+in+local+time

## Links

* Minutes Google Doc: <https://docs.google.com/document/d/1UFy4N1DrZ00XnVK-DsnrCVHXwcpkawjs2VxPu8UJfv8/edit>

## Agenda

Extracted from **cross-project-council-agenda** labelled issues and pull requests from the **openjs-foundation org** prior to the meeting.

### openjs-foundation/summit

* Travel Fund Questions for Project Community Members [#249](https://github.com/openjs-foundation/summit/issues/249)

### openjs-foundation/cross-project-council

* \[Action Requested\] Nominations Sought for 2020-2021 Voting CPC Membership  [#488](https://github.com/openjs-foundation/cross-project-council/issues/488)
* Projects with Windows and Mac installers? [#476](https://github.com/openjs-foundation/cross-project-council/issues/476)
[...]

PR #509, opened 2020-03-31
doc: adding meeting notes March 17 24 31

joesepi:
closes: #493
closes: #500
closes: #506


kborchers:
I can approve the 3/31 notes but I was on vacation for the 3/17 and 3/24 meetings and not present so I will abstain from approving those notes.

PR #515, opened 2020-04-07
[WIP] Adds reporting emails 

jorydotcom:
This PR adds a list of the project-specific CoC reporting emails to our code of conduct, closing the loop for projects that wish to reference our code of conduct by providing a pointer to the URL https://code-of-conduct.openjsf.org.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -11,11 +11,18 @@ Refer to the sections on reporting and escalation in this document for the speci
 
 ### Project Spaces
 
-For reporting issues in spaces related to a member project please use the email provided by the project for reporting. Projects handle CoC issues related to the spaces that they maintain. Projects maintainers commit to:
+For reporting issues in spaces related to a member project, please use the email provided by the project for reporting which has been provided below. Projects handle CoC issues related to the spaces that they maintain. Projects maintainers commit to:
 
 - maintain the confidentiality with regard to the reporter of an incident
 - to participate in the path for escalation as outlined in
   the section on Escalation when required.
+
+#### Project Reporting Emails
+If a project has not provided a reporting path, you may use the escalation email.
+
+* AMP Project - `amphtml-coc@googlegroups.com`
+* Electron - `coc@electronjs.org`
+* Node.js - `report@nodejs.org`
 
 ### Foundation Spaces
 
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -18,7 +18,6 @@ For reporting issues in spaces related to a member project, please use the email
   the section on Escalation when required.
 
 #### Project Reporting Emails
-If a project has not provided a reporting path, you may use the escalation email.
 
 * AMP Project - `code-of-conduct@amp.dev`
 * Electron - `coc@electronjs.org`
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -20,8 +20,34 @@ For reporting issues in spaces related to a member project, please use the email
 #### Project Reporting Emails
 
 * AMP Project - `code-of-conduct@amp.dev`
+* Appium - [email maintainers]
+* Architect - 
+* Dojo - [reporting form](https://docs.google.com/forms/d/1ZiCUv3FsarWetoQ8J5QnCdIrHLj0xMmwNSLCyBkio7c/viewform?edit_requested=true)
 * Electron - `coc@electronjs.org`
+* ESLint - 
+* Esprima - 
+* Express.js - `express-coc@lists.openjsf.org`
+* Fastify - `hello@matteocollina.com` or `tommydelved@gmail.com`
+* Grunt - 
+* HospitalRun - `hello@hospitalrun.io`
+* Interledger.js - 
+* Intern - 
+* JerryScript - 
+* jQuery - 
+* libuv - 
+* LoDash - 
+* messageformat - 
+* Marko - 
+* Mocha - `report@lists.openjsf.org`
+* Moment.js - 
 * Node.js - `report@nodejs.org`
+* Node-RED - `team@nodered.org`
+* nvm - 
+* Pep - 
+* QUnit - 
+* Webdriver.io - [contact TSC reps](https://github.com/webdriverio/webdriverio/blob/master/AUTHORS.md)
+* Webhint - `support@webhint.io`
+* webpack - [redirects here]
 
 ### Foundation Spaces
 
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -11,12 +11,19 @@ Refer to the sections on reporting and escalation in this document for the speci
...
 - maintain the confidentiality with regard to the reporter of an incident
 - to participate in the path for escalation as outlined in
   the section on Escalation when required.
 
+#### Project Reporting Emails
+If a project has not provided a reporting path, you may use the escalation email.
+
+* AMP Project - `amphtml-coc@googlegroups.com`
+* Electron - `coc@electronjs.org`
+* Node.js - `report@nodejs.org`
```
mcollina, line 25: I’m not in agreement with this whole paragraph. The CoCP is not a moderation path for things that can should the projects directly.

This list is also hard to maintain. I think the reporting email should be in the CoC of the project (and possibly checked during onboarding).

jorydotcom: The issue is that we have a circular dependency. We require projects to adopt the Foundation's Code of Conduct, which many of them will incorporate by directly linking to our document (this one, hosted at code-of-conduct.openjsf.org). When a person then gets to the section on reporting, they are directed back to the project for reporting information, which points them back here. So they are more likely to report to our escalation path than to the project's specific escalation path. I agree that the project should make its own reporting emails visible. 

I disagree that this will be hard to maintain. Reporting emails for each project won't change frequently, if at all. This is the simplest available solution for projects to adopt the foundation code of conduct uniformly while we work out the details.

brianwarner: As an option: we can host projects' CoC lists on lists.openjsf.org, which would have the benefit of a consistent alias pattern (e.g., express-coc@, amp-coc@, messageformat-coc@, etc), in addition to being consistent long term, per Jory's observation.

mhdawson: > If a project has not provided a reporting path, you may use the escalation email.

is the part that concerns me. I don't think we've agreed the CPC we act as the moderation first point of contact for projects that don't want to do it themselves (I might be mistaken on this please chime if that's the case).  My personal feeling is that a requirement for a project being part of the Foundation is that they will have a reporting path. Reports can be escalated if the project does not handle them properly but that occurring regularly should trigger some sort of action beyond just handling the escalations.

CODE_OF_CONDUCT.md
```diff
@@ -11,12 +11,19 @@ Refer to the sections on reporting and escalation in this document for the speci
...
+For reporting issues in spaces related to a member project, please use the email provided by the project for reporting which has been provided below. Projects handle CoC issues related to the spaces that they maintain. Projects maintainers commit to:
 
 - maintain the confidentiality with regard to the reporter of an incident
 - to participate in the path for escalation as outlined in
   the section on Escalation when required.
 
+#### Project Reporting Emails
+If a project has not provided a reporting path, you may use the escalation email.
+
+* AMP Project - `amphtml-coc@googlegroups.com`
```
tobie, line 23:
```suggestion
* AMP Project - `code-of-conduct@amp.dev`
```

CODE_OF_CONDUCT.md
mcollina: @mhdawson captured my objection. I’m marking it explicitly.

molant: Related items in the graduation checklist are:

> * Adopt the OpenJS Foundation Code of Conduct
> * Update project CoC reporting methods to include OpenJS Foundation escalation path

With the second item we assume there's already a way to report in the project. Maybe we should double check all projects have one and update the phrasing here?

bnb: >As an option: we can host projects' CoC lists on lists.openjsf.org, which would have the benefit of a consistent alias pattern (e.g., express-coc@, amp-coc@, messageformat-coc@, etc), in addition to being consistent long term, per Jory's observation.

I am an explicit -1 on centralizing this into the Foundation's control. I would be fine with providing this as an option, but I would genuinely prefer that we explicitly do not require this of projects if they already have their own processes or do not want to leverage this.

bnb: >I don't think we've agreed the CPC we act as the moderation first point of contact for projects that don't want to do it themselves (I might be mistaken on this please chime if that's the case).

@mhdawson I generally agree with what you said, but it is worth noting that the current Onboarding process perhaps optimistically requires that the escalation path be included for new projects coming in. If this is not something that's completely agrees upon, it should not be in the onboarding.

mhdawson: I think we've agreed that including the escalation path is required... But for escalation, not as the default method of reporting a CoC related to projects.

jorydotcom: I have removed the sentence that suggested projects without a reporting path can use the CoCP email.

MylesBorins: As it is currently updated I think we should go ahead and just land this. It is important documentation and not having it is creating actual risk of people not know how / where to make reports. I'd like to request that we land this now and iterate to remove it later if we find maintaining it is an issue

jorydotcom:
For additional context, I've just pushed a commit that documents the available reporting addresses for Impact, Growth, At Large and Incubating projects. As you can see there are ~16 projects with no explicit reporting email address. Projects fall into a number of cases:

1. The CoC reference points to this document, but no reporting address is provided.
2. The Code of Conduct file or reference was not found.
3. The Code of Conduct file points to the old JS Foundation Code of Conduct, or a corporate contributor's CoC.
4. A CoC file is present but the reporting path is not.

IMO items 2-4 above can and should be addressed as each project completes its onboarding tasks; this PR would provide the lowest friction way for smaller projects to keep their CoCs up-to-date and make sure their reporting paths are discoverable.

CODE_OF_CONDUCT.md
mcollina: This is ok now

CODE_OF_CONDUCT.md
```diff
@@ -11,12 +11,44 @@ Refer to the sections on reporting and escalation in this document for the speci
...
+* ESLint - 
+* Esprima - 
+* Express.js - `express-coc@lists.openjsf.org`
+* Fastify - `hello@matteocollina.com` or `tommydelved@gmail.com`
+* Grunt - 
+* HospitalRun - `hello@hospitalrun.io`
+* Interledger.js - 
+* Intern - 
+* JerryScript - 
+* jQuery -
```
dmethvin, line 36:
```suggestion
* jQuery - `security@jquery.com`
```

jorydotcom:
Transferring this issue to the new / code-of-conduct repo

PR #544, opened 2020-05-12
make it clearer there are two proposals

MylesBorins:
small editorial change


brianwarner:
LGTM. As this is an editorial change to avoid confusion, I recommend fast tracking.

PR #545, opened 2020-05-12
Add "Reason for Applying" to new project template

eemeli:
Projects of various sizes apply to join the foundation for a variety of reasons, and they hold a variety of expectations on what membership of the foundation may bring them. Most of these are entirely valid, but we can't always know that.

Hence I think we should directly ask applicants why they'd like to join, and what they expect to benefit from their membership. This should make it much easier for the CPC to evaluate applications, and for any misunderstandings to be corrected as early as possible.

### DIFF
--- a/NEW_PROJECT_APPLICATION.md
+++ b/NEW_PROJECT_APPLICATION.md
@@ -8,6 +8,10 @@ Who is the main point of contact during the application process?
 
 A rough description of the project in less than 100 words.
 
+## Reason for Applying
+
+A short description of the reasons why your project is applying to join the foundation, and your expectations from the foundation.
+
 ## Statement of alignment with OpenJS Foundation charter and mission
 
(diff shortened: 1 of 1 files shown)


NEW_PROJECT_APPLICATION.md
```diff
@@ -8,6 +8,10 @@ Who is the main point of contact during the application process?
 
 A rough description of the project in less than 100 words.
 
+## Reason for Applying
+
+A short description of the reasons why your project is applying to join the foundation, and your expectations from the foundation.
```
tobie, line 13: Specifying a word count would be useful.

eemeli: Any suggestions for a number? I'm really rather ambivalent on this. React to this with 🚀for 250, 👀for 100, or 👎to keep it unspecified.

tobie: Having gone through the process myself, having a clear expectation of size was incredibly helpful.

eemeli: Based on feedback, I added a limit of 100 words.

PR #558, opened 2020-05-28
Add language that former Voting members become Regular members when t…

brianwarner:
Add language that former Voting members become Regular members when their terms end.

This doesn't seem to change the definition of Regular or Voting members in the Charter, so I limited the changes to the README and GOVERNANCE files.

### DIFF
--- a/GOVERNANCE.md
+++ b/GOVERNANCE.md
@@ -1,6 +1,6 @@
 # OpenJS Cross Project Council
 
-For the current list of Team members, see the project [README.md][README].
+For the current list of Team members, see the project [README.md][README] and the [OpenJS Foundation Directory](https://github.com/openjs-foundation/directory-private/blob/master/cpc-private.md).
 
 ## Members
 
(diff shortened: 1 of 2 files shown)


GOVERNANCE.md
```diff
@@ -1,6 +1,6 @@
 # OpenJS Cross Project Council
 
-For the current list of Team members, see the project [README.md][README].
+For the current list of Team members, see the project [README.md][README] and the [OpenJS Foundation Directory](https://github.com/openjs-foundation/directory-private/blob/master/cpc-private.md).
```
tobie, line 3: I suggest we use as a rule to always mark private information as such and give the reason as to why it is private. 

```suggestion
For the current list of Team members, see the project [README.md][README] and the [OpenJS Foundation Directory](https://github.com/openjs-foundation/directory-private/blob/master/cpc-private.md) (restricted access for privacy reasons).
```

brianwarner: This looks good, thanks @tobie.

PR #489, opened 2020-03-10
Security reporting proposal (stage 0)

MarcinHoppe:
I wrote a stage 0 proposal in response to #326. The immediate goal is to get feedback about the level of support the Foundation can offer to individual projects and how consistent we want the reporting mechanism to be across projects (it varies today).

Feedback and comments will be very much appreciated.

PR #566, opened 2020-06-16
Move security reporting proposal to stage 2

MarcinHoppe:
This PR moves the security reporting proposal to stage 2, per [agreement](https://github.com/openjs-foundation/cross-project-council/pull/489#issuecomment-641597486) in the last CPC meeting.


mhdawson:
I think this looks good, we should probably just get some project feedback before approving.

designMoreWeb:
LGTM

mhdawson:
@jorydotcom do you want to send out the email to ask projects for input or should I do that directly.  Making sure projects don't have any concern with respect to the new requirements outlined is the current blocker.

jlipps:
👍 from appium

jorydotcom:
> @jorydotcom do you want to send out the email to ask projects for input or should I do that directly. Making sure projects don't have any concern with respect to the new requirements outlined is the current blocker.

sent!

MarcinHoppe:
@ljharb is this something we can put as a hard requirement, or is this something that we want to encourage? I am not all that familiar with Foundation projects to see if every project has its own GitHub org.

mhdawson:
@MarcinHoppe, @ljharb how about the recommendation being adding the security policy to the org's .github repo if the project has an org and adding to the repo otherwise?

nzakas:
> I prefer the approach of adding the security policy to the org's .github repo, so that it shows up in $repoURL/security/policy automatically, both without needing to clutter individual repos, and also with no risk of forgetting to add it.

I agree with @ljharb. For projects with multiple repos in an organization, it makes much more sense to have a central location for the security policy. 

As an addition, I’d recommend requiring a link to the security policy in each project’s README. That way, the information is always front-and-center even if the actual policy lives elsewhere.

ljharb:
@MarcinHoppe When there's no org, you can have a user-wide `.github` repo that works identically.

ljharb:
@nzakas every repo has a "security" tab, above all the content, whose "policy" subtab would contain that info. I'm not sure having it in the (often long) readme adds value.

jaskiratsingh2000:
Good approach.

nzakas:
@ljharb The security tab is helpful when people are viewing the project on GitHub but not if they are viewing it on npm. The README ends up being copied to npm so you end up with a link everywhere the README is found.

ljharb:
That's a good point. I wonder if (separately) we should ask npm about adding a field to package.json for the security policy that they could prominently display on a package page?

nzakas:
That would be great! :+1:

MarcinHoppe:
@ljharb @nzakas @mhdawson I added a sentence to recommend org-wide deployment of the security policy file.

mhdawson:
We agreed in the CPC meeting today that we'd wait until the end of the week to see if there is any more feedback from the projects. If not it can land at that point.

mhdawson:
No additional comments so landing as per earlier agreement.

PR #593, opened 2020-07-20
Moving Supporter Program to Phase 1

SaraJo:


proposals/stage-1/SUPPORTER_PROGRAM/README.md
```diff
@@ -0,0 +1,48 @@
...
+[Sara Chipps](https://github.com/sarajo)
+ 
+## Description
+
+Membership program for OpenJS individual supporters. 
+
+## Required Resources
+
+What resources would be needed to execute it?
+* The current weekly newsletter signup is removed from the OpenJS website.
```
mcollina, line 15: I think the newsletter should stay open to all project collaborators and members.

SaraJo: That makes sense. Is there another offering we can give supporters that will incentivise people to join?

.gitignore
```diff
@@ -1 +1,3 @@
 node_modules/
+*DS_Store
+.DS_Store
```
MylesBorins, line 3: get out of here DS_STORE amirite?

proposals/README.md
```diff
@@ -14,6 +14,8 @@ Anyone may submit an idea for a policy or program following the [staging process
 
 * [Charter Review](./stage-1/CHARTER_REVIEW)
 * [Collab Summit](./stage-1/COLLAB_SUMMIT)
+* [Collaboration Network](./stage-0/COLLABORATION_NETWORK)
```
MylesBorins, line 17: I think this might be left over from rebase

```suggestion
```

proposals/stage-1/SUPPORTER_PROGRAM/README.md
MylesBorins: Is this the same newsletter as the one we use for projects? My guess is this is a separate marketing based one.

If we offer free supporter status to contributors that could be a path for them to gain acess.

Are we going to stop supporting a public mailing list altogether here? Seems like that could be a step backwards for some marketing efforts such as advertising the conference.

proposals/stage-1/SUPPORTER_PROGRAM/README.md
```diff
@@ -0,0 +1,48 @@
...
+## Required Resources
+
+What resources would be needed to execute it?
+* The current weekly newsletter signup is removed from the OpenJS website. 
+    * Existing subscribers will not be removed
+    * As people sign up to be an individual supporter they will be added to the mailing list
+    * They can request removal at any time
+    * Optional additional email content, such as “What is the CPC and how can I get involved” can be set up via an email drip campaign for new supporters. 
+* Design
+    * OpenJS Member badges, avatars, and swag
```
MylesBorins, line 21: where would badges / avatars lives? Would there be a social network or profile page?

proposals/stage-1/SUPPORTER_PROGRAM/README.md
SaraJo: My understanding is that right now it has a little over a hundred subscribers. The suggestion was to use the newsletter that Jory send out as a supporter newsletter. 

If we're not in agreement about that then would love to brainstorm a new newsletter (and maybe figure out who would write/distribute it)? Or, something else we could offer that would incentivise people to sign up?

proposals/stage-1/SUPPORTER_PROGRAM/README.md
SaraJo: I was thinking we could give it to them for their social media profiles/websites. WDYT?

proposals/stage-1/SUPPORTER_PROGRAM/README.md
MylesBorins: TBH I don't have particularly strong feelings about this. We also have twitter and other social channels to send out comms... just want to ensure that we are not losing outreach channels

proposals/stage-1/SUPPORTER_PROGRAM/README.md
MylesBorins: Seems reasonable to me.

Before stage 2 we will likely want to make a more in depth proposal about what site that is, and how profiles will work.

proposals/stage-1/SUPPORTER_PROGRAM/README.md
SaraJo: We discussed this at the CPC meeting, it will stay open AND be a perk.

proposals/stage-1/SUPPORTER_PROGRAM/README.md
SaraJo: Yes, that makes lots of sense.

Issue #326, opened 2019-09-18
Responsible security disclosures


tobie:
Is there a standard way for OpenJSF projects to accept responsible security disclosures or is the expectation that projects set up their own solution?

Node.js has its [own policy](https://nodejs.org/en/security/), for example. I don't know about other projects.

I feel like it would be nice if this was something the foundation handled for projects (even if the project would be the one responsible for dealing with the disclosure itself).

wesleytodd:
Related: https://github.com/nodejs/package-maintenance/issues/159

Our Node WG has been having some discussions around this.  If the foundation was going to have something it should align with what the Maintenance WG and Security WG recommend.

MylesBorins:
We've definitely discussed trying to replicate some of the work that has
been done in core... especially around managing CNA and CVE stuff.

With githubs new announcement though perhaps we should look at working with
them... as the management of the CNA is a non-trivial amount of work.

mhdawson:
The Node.js security work also does work in this area, and HackerOne who we work with handle CVE stuff for the ecosytem repors.  I've pinged that team through https://github.com/nodejs/security-wg/issues/578

mhdawson:
There have been discussions as to wether part of the Node.js security-wg which focuses on ecosystem triage should move up to the OpenJS level.  This is probably a good time to continue that discussion.  @sam-github

sam-github:
https://github.com/nodejs/security-wg/pull/579 relates to this slightly, in that if the responsibilities of the nodejs/security-wg become more clearly and singularly focussed on the Ecosystem outside Node.js, they might be a candidate to move up a level.

Also, note that even though many, perhaps most, of the vulns in the DB are reported against "Node.js" packages, a number of those packages can also run in the browser, and there are at least a handful (or were, last time I surveyed them a couple years ago) that are npm package that run ONLY in the browser, so the WG is already in some sense collecting and triaging non-Node.js vulnerabilities.

MarcinHoppe:
I brought this up some weeks ago when triaging a report of a vulnerability in a React component: clearly only intended to work in the browser but still distributed via the npm.

I would definitely be up for discussion about taking our work in the Node.js ecosystem triage team to the OpenJS foundation level.

mhdawson:
Created https://github.com/nodejs/security-wg/issues/589

mhdawson:
PR in package main repo which is related: https://github.com/nodejs/package-maintenance/pull/277

[...]

PR #602, opened 2020-07-29
Move security reporting proposal to stage 3

MarcinHoppe:
Moving the security reporting proposal to stage 3, per discussion in #326.


mikesamuel:
The difference between stage 2 and 3 is at [STAGING_PROCESS](https://github.com/openjs-foundation/cross-project-council/blob/master/STAGING_PROCESS.md#stage-2---accepted-proposal)

(Emphases added)

> ## Stage 2 - Accepted Proposal
> Proposals in this stage are nearly complete. They may be in a pilot or trial-run mode for a limited audience, or pending legal review. They may be awaiting budget approval. The purpose of this stage is to **complete any operational checks, such as acquiring budget, access, or Board approval if needed, before launching the program**.

Is the bolded part primarily what we need to discuss?

> ## Stage 3 - Complete
> The proposal is now an adopted policy or program of the Foundation, and is owned by the committee or role identified. It is expected that the policy or program **will be re-evaluated on some regular basis and deprecated if need be**.

IIUC, stage 3 does not complicate revision based on unforeseen needs.

mhdawson:
@MarcinHoppe in this stage we should move the files out of the proposal and into their final locations in the repo.  I think that means finding the right place/name for REQUIREMENTS.md in the Cross Project Council Repo or Onboarding repo and updating other docs as appropriate to reference it (For example I think it should be referenced in the onboarding doc as one of the requirements)

MarcinHoppe:
@mikesamuel @mhdawson Great feedback, I will work on that.

MarcinHoppe:
Sorry for a little delay here but I've been on vacation. I'll move this forward before the next CPC meeting.

MarcinHoppe:
Closing in favour of #633.

PR #619, opened 2020-08-13
Ads links to CPC members' GitHub profiles

tobie:
This humanizes the CPC.

It also makes it easier to find out how to contact individual members, while still leaving it up to the individual members to decide how much contact information they want to provide through their Github profile.


eemeli:
Recommend this be fast-tracked.

PR #620, opened 2020-08-15
License preference, per discussion with legal

brianwarner:
From a discussion on the Board, I checked with OpenJS Foundation legal counsel on the preferred license, all other things being equal. While we have a number of options in the IP policy, Apache 2.0 seems to be the preferred option if there are no other reasons to choose something else.

### DIFF
--- a/proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
+++ b/proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
@@ -73,7 +73,7 @@ The Application Champion is a member of the CPC or Foundation staff who commits 
 ##### Initial fit checklist
 
 1. Are the existing members/space leaders agreed to managing the space under open governance?
-1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)?
+1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)? Where possible, it is the Foundation's preference that new work/artifacts be licensed Apache 2.0.
 1. Have the existing members/space leaders agreed that the collaboration space will adopt the [OpenJS CoC](https://code-of-conduct.openjsf.org)?
 1. Are licences for any existing material that will be added to the collaboration space repo compatible with Foundation requirements?
 1. Fit with OpenJS Foundation (not all of these need to be met, but provide different ways of assessing strategic fit).
--- a/proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
+++ b/proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
@@ -73,7 +73,7 @@ The Application Champion is a member of the CPC or Foundation staff who commits 
 ##### Initial fit checklist
 
 1. Are the existing members/space leaders agreed to managing the space under open governance?
-1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)? The IP policy permits projects to use a variety of licenses, including MIT, BSD-2, BSD-3, and Apache 2.0. Where there is no preference otherwise, the Foundation recommends that new work/artifacts be licensed Apache 2.0.
+1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)? The IP policy permits projects to use a variety of licenses, including MIT, BSD-2, BSD-3, and Apache 2.0. Where there is no preference otherwise, the Foundation recommends that new work/artifacts be licensed Apache 2.0 due to its explicitly defined rights and responsibilities.
 1. Have the existing members/space leaders agreed that the collaboration space will adopt the [OpenJS CoC](https://code-of-conduct.openjsf.org)?
 1. Are licences for any existing material that will be added to the collaboration space repo compatible with Foundation requirements?
 1. Fit with OpenJS Foundation (not all of these need to be met, but provide different ways of assessing strategic fit).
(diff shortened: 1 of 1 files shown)


proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
```diff
@@ -73,7 +73,7 @@ The Application Champion is a member of the CPC or Foundation staff who commits
 ##### Initial fit checklist
 
 1. Are the existing members/space leaders agreed to managing the space under open governance?
-1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)?
+1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)? Where possible, it is the Foundation's preference that new work/artifacts be licensed Apache 2.0.
```
ljharb, line 76: can i ask why? MIT is the *resounding* majority in the JS ecosystem; apache-2 is quite rare there (it's more common in python)

ljharb:
In other words, I'd say for a foundation purporting to be for JS, "not MIT" is to an *exceeding* degree what needs the burden of proof.

proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
eemeli: My presumption would be that it's due to the explicit patent grants included in Apache 2, unlike the other supported licenses. But this definitely does need some clear explanation around it arguing the case.

brianwarner: That's right on the explicit patent grants, @eemeli. 

That said, I would actually recommend against arguing any case here. The idea here was not to get too prescriptive, but rather to provide gentle guidance to someone who genuinely doesn't have an opinion, so they don't feel they are choosing from random. If there's even the slightest preference otherwise, such as what you're saying @ljharb, then they're definitely empowered by the IP policy to choose what they prefer. But since that's already written down (and following @tobie's general lead on keeping things simple) I'm not sure it makes sense to get into it here.

Thoughts?

dylans: > can i ask why? MIT is the _resounding_ majority in the JS ecosystem; apache-2 is quite rare there (it's more common in python)

My experience is that in general lawyers and enterprises prefer Apache 2, with a second choice being the new BSD, and a third choice being MIT. Generally the reverse order is true for JS open-source maintainers. For most practical purposes the licenses are roughly the same and generally considered equal, except that Apache is explicit around patent rights.

The primary reason JS devs historically chose MIT over the years is that it is terse, and in the early days the necessary approach was to require the license at the top of each file. The MIT license was short enough to do that, whereas the Apache license was a few pages and back before bundlers/build tools were popular, embedding the license would add a significant increase to file size. Either that or it's a function of tl;dr;. And from there as an ecosystem we typically just copy the license used by other projects we respect.

tobie: This feels like an odd context to express that preference. If the Board/Foundation feels strongly about Apache 2 over MIT, why not to some advocacy and evangelism on the blog and share the arguments behind the suggestion? This feels a bit underhanded to me.

ljharb: It’s also not a universal preference - the lawyers at a previous employer did not prefer Apache over mit at all, and in fact their strong preference was “do whatever the relevant ecosystem prefers”.

brianwarner: I'd just maybe recommend we take a step back, as the point was not that the board feels strongly about choosing between one or another pre-approved license. The discussion really was just "If a contributor is creating something new and doesn't have a license preference, which of the four options would be recommended so they aren't choosing at random?" The key thing is that this just establishes a default for this particular program, and if there's any reason to choose another license they're able to do so under the IP policy (as with OpenJS Foundation projects).

The main takeaway is that the board felt this was a missing bit of info, so the action was to check with counsel and add a line to the proposal so they can vote on it.

ljharb: Suggesting a default makes total sense :-) i commented because i think suggesting any default for a JS project besides "MIT" is confusing, since MIT (or ISC) is by a *large* margin the most common license used in the JS ecosystem.

proposals/stage-2/COLLABORATION_NETWORK/COLLABORATION_SPACE_PROGRESSION.md
```diff
@@ -73,7 +73,7 @@ The Application Champion is a member of the CPC or Foundation staff who commits
 ##### Initial fit checklist
 
 1. Are the existing members/space leaders agreed to managing the space under open governance?
-1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)?
+1. Have the existing members/space leaders agreed that all work will happen under the [OpenJS Foundation IP Policy](https://ip-policy.openjsf.org)? The IP policy permits projects to use a variety of licenses, including MIT, BSD-2, BSD-3, and Apache 2.0. Where there is no preference otherwise, the Foundation recommends that new work/artifacts be licensed Apache 2.0.
```
ljharb, line 76: it'd be great if this included *why* the foundation recommends that.

MylesBorins (approved):
LGTM

kborchers (approved):

tobie (approved):
I like the language in that last iteration. It is much better. I still believe this would be better spelled out in a broader context (maybe the IP policy), but I don't want to block on this.

mhdawson (approved):
LGTM

sendilkumarn (approved):

dmethvin (approved):

eemeli (approved):

molant (approved):

joesepi (approved):

PR #633, opened 2020-09-01
Move security reporting proposal to stage 3

MarcinHoppe:
Moving the security reporting proposal to stage 3, per discussion in #326.

The corresponding checklist update: openjs-foundation/project-onboarding#52.

This PR supersedes #602.


mikesamuel (approved):

kborchers (approved):

timmywil (approved):

joesepi (approved):

MylesBorins:
Added `board approval required` because that is usually a gate for stage-3. At the very least an FYI seems appropriate. If this has already been discussed and there was consensus we don't need anything from the board please feel free to dismiss

MarcinHoppe:
@MylesBorins I don't think this was ever discussed, so thanks for flagging this PR.

mhdawson:
@Myles I think we'd mentioned that we did not think it needed board approval as its not chaning the charter of the CPC.  Not that I'm opposed. I do think we've mentioned that it was underway already.

MylesBorins:
@mhdawson That seems reasonable to me... I guess if there is no need for board approval then 2 -> 3 should basically be automatic

tobie (approved):

ljharb (approved):

mikesamuel (approved):

PR #650, opened 2020-09-22
Remove Growth Stage

tobie:
This is a draft proposal to remove the Growth Stage and add a Growth plan instead.

Worth noting that this has broad implication in the whole CPC repository which still needs to be tackled.


PROJECT_PROGRESSION.md
```diff
@@ -252,7 +224,18 @@ It is expected that incubating project will make an active effort to work throug
 
 1. consensus within private mailing list to move into incubation process
 
-## IV. Annual Review Process
+## IV. Growth Plan
+
+At Large project that are interested in reaching Impact Stage need to develop a growth plan to so do. This plan needs to be developed in collaboration with the CPC and approved by it.
```
mhdawson, line 229: I think in the discussions, the idea was that a Growth plan is not tied to wanting to reach Impact Stage, and that even Impact projects might want a Growth plan.

Maybe

Project that are interested in extra support/attention in order to grow in a certain area need to developer a plan to do so....

eemeli:
If we're going to drop "Growth", I think the language we use about projects needs to change more than what's suggested here. In particular, I don't think it makes sense any more to talk of project "stages". As in, we have projects, and some of them we've identified/marked as having sufficient impact on the whole ecosystem to get a greater voice in the foundation as well.

Put another way, I think we're talking about a move from:
```ts
interface Project {
  type: 'at-large' | 'growth' | 'impact' | 'emeritus'
  get cpcImpact() { return this.type === 'impact' }
  ...
}
```
to:
```ts
interface Project {
  active: boolean
  impact: boolean
  get cpcImpact() { return this.active && this.impact }
  ...
}
```

PROJECT_PROGRESSION.md
tobie: Agreed that what we discussed.

I'm curious however, what would the incentives be for a project to go through the foundation's structure for this? Why not just create their own roadmap?

Projects that are doing this to move to Impact stage are incentivized by the status/recognition aspect and additional CPC weight.

I think I'm inclined to suggest starting off with this a just a growth plan for At Large to Impact at first (as an MVP, if you will) and opening it up if there's demand.

tobie:
I get your point @eemeli, but I also think that the stages in terms of maturity makes sense. Maybe we should reframe the stage names a bit to match that more?

tobie:
Following our discussion today, I feel like this is the MVP we need.

It:

- removes growth as a stage,
- creates a growth plan.

Concretely, this would require moving all existing Growth projects to At Large, and having projects that want to move to impact just create a growth plan in the `project-status` repo.

PROJECT_PROGRESSION.md
joesepi: I think I agree with @mhdawson here. My understanding is that any project, even an Impact project, could adopt a growth plan. It could be that the growth plan is for an At-Large project to become an Impact project, but it also could just be to grow contributors, to reach a milestone, or a variety of any other reasons. Perhaps even moving to Emeritus could be considered a "growth plan". ¯\_(ツ)_/¯

tobie:
In the CPC meeting today, we thought it would be good to ask Growth Projects what they thought about having Growth stage removed. @jorydotcom, what would be the best way of going about that?

PROJECT_PROGRESSION.md
```diff
@@ -203,13 +173,15 @@ Impact Stage projects are expected to participate actively in CPC proceedings, a
 
 #### Acceptance Criteria
 
-To graduate from At Large or Growth status, or for a new project to join as an Impact project, a project must meet the Growth stage criteria plus:
+To graduate from At Large stage or for a new project to join as an Impact project, a project must meet the following criteria:
 
+ * Document that it is being used successfully in production by at least two independent end users which, in the CPC’s judgement, are of adequate quality and scope.
  * Have a defined governing body of at least 5 or more members (owners and core maintainers), of which no more than 1/3 is affiliated with the same employer. In the case there are 5 governing members, 2 may be from the same employer. 
- * Have a documented and publicly accessible description of the project's governance, decision-making, and release processes.
+ * Have a documented and publicly accessible description of the project's governance, decision-making, contribution, and release processes. This is preferably laid out in a GOVERNANCE.md file and references a CONTRIBUTING.md and OWNERS.md file showing the current and emeritus committers.
```
christian-bromann, line 180: I would suggest using `AUTHORS.md` given `OWNERS.md` can be confused with `OWNERS` which is not for documentation purposes as it has an effect on how GitHub handles certain things in the repository.
```suggestion
 * Have a documented and publicly accessible description of the project's governance, decision-making, contribution, and release processes. This is preferably laid out in a GOVERNANCE.md file and references a CONTRIBUTING.md and AUTHORS.md file showing the current and emeritus committers.
```

christian-bromann (commented):
Besides becoming a CPC voting member are there any other incentives a project would be interested in to transition from At Large to Impact project?

boneskull:
The "growth" stage represents a path for a project to reach the impact stage.   Within the growth stage, a project should work with the CPC on a mutually-agreed-upon set of goals which will satisfy the requirements to become an impact project. **If a growth stage project hits those goals, it graduates to the impact stage.**

**I really want to make sure everybody understands the above.**  If you _don't_ see it this way, or feel like hitting all the growth plan goals should _not_ automatically move a project into the impact stage, then we need to discuss.  Maybe we're already all on the same page here, but I can't tell.

That said, I don't feel strongly about removing the stage itself.  But I also don't understand why we shouldn't keep it.  Why do this now?  Are there negative consequences to just leaving it as-is?

boneskull:
Regarding the "any project can have a growth plan" thing: if you're an impact project, there's no point in working on a growth plan with the CPC.  I would argue that working on a growth plan with an impact project is just not a good use of our time.  When we're talking about growth plans, we need to prioritize the projects who are doing it with an eye towards impact stage.

jorydotcom:
@boneskull I think it's true that 
* some projects will use the 'growth stage' to achieve 'impact stage' and that should be prioritized
* some 'at large stage' projects need extra resources to achieve a goal, but won't need or want to be impact projects
* some 'impact stage' projects _may_ want to scope a goal they can achieve with the CPC

With this in mind, I feel like the 'Growth Stage' is really a different sort of program that we can/should better define - e.g. Projects electing to create a growth plan can request $x budget or y tool or z personnel resources to help reach their goal of (becoming an impact project, creating a mentorship program, expanding/diversifying a maintainer pool, improving documentation etc).

boneskull (approved):

joesepi:
@tobie will check with AMP team next week and then we will move forward.

tobie:
> @tobie will check with AMP team next week and then we will move forward.

Done, see https://github.com/ampproject/meta-tsc/issues/55. AMP's TSC is totally fine with removing the growth stage.

PROJECT_PROGRESSION.md
```diff
@@ -203,13 +173,15 @@ Impact Stage projects are expected to participate actively in CPC proceedings, a
...
- * Have a documented and publicly accessible description of the project's governance, decision-making, and release processes.
+ * Have a documented and publicly accessible description of the project's governance, decision-making, contribution, and release processes. This is preferably laid out in a GOVERNANCE.md file and references a CONTRIBUTING.md and AUTHORS.md file showing the current and emeritus committers.
+ * Demonstrate a substantial ongoing flow of commits and merged contributions.
  * Have a healthy number of committers from at least two organizations. A committer is defined as someone with the commit bit; i.e., someone who can accept contributions to some or all of the project.
  * Adopt the Foundation Code of Conduct.
- * Explicitly define a project governance and committer process. This is preferably laid out in a GOVERNANCE.md file and references a CONTRIBUTING.md and OWNERS.md file showing the current and emeritus committers.
+ * Accept security reports and publicly disclose them after the fix has been made available.
```
eemeli, line 184: I'm going to presume that we would not use a lack of security issues as a blocker for Impact, provided that the policies are in place for dealing with such. While it's of course likely that something like that will be encountered by a project before it reaches this level, I could well imagine such issues having all been caught at an early enough stage to not need to follow any special procedures for dealing with them.

I don't think this language needs to change, I just wanted to note here that this exact verbiage should not be understood to require a security report to have been made in the project.

eemeli (approved):

PROJECT_PROGRESSION.md
eemeli: Also agree with @mhdawson. As I see it, a Growth Plan is a tool for which we presume that the most use will be for projects wanting to transition from At-Large to Impact, but there should not be a built-in limitation preventing that tool from being used for any other purpose, which we might not be able to imagine or exhaustively enumerate at this time.

PROJECT_PROGRESSION.md
mhdawson: I read it as the project needs to have a process to accept security reports and handle them but no requirement that the process have been exercised.

PROJECT_PROGRESSION.md
tobie: I like the idea of having a plan for each shift:

* Incubating -> At Large/Impact
* At Large -> Impact
* At Large/Impact -> Emeritus

They're different plans, though. I'm also less sure about the relevance of having a generic growth plan. Has anyone asked for this? Why don't we get there once someone actually asks?

PROJECT_PROGRESSION.md
```diff
@@ -203,13 +173,15 @@ Impact Stage projects are expected to participate actively in CPC proceedings, a
 
 #### Acceptance Criteria
 
-To graduate from At Large or Growth status, or for a new project to join as an Impact project, a project must meet the Growth stage criteria plus:
+To graduate from At Large stage or for a new project to join as an Impact project, a project must meet the following criteria:
 
+ * Document that it is being used successfully in production by at least two independent end users which, in the CPC’s judgement, are of adequate quality and scope.
```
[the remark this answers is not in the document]
tobie: To be clear, this was pre-existing text. I'm just reorganizing what's already here.

dylans (approved):

eemeli (approved):

mhdawson (approved):
LGTM

PR #653, opened 2020-10-02
moved supporter program to stage 3

SaraJo:


mhdawson:
For the PR to move to stage 3, I think we should have the files which will end up living in the non-proposal sections of the CPC repo, for example should the SUPPORTER_PROGRAM.md or a file with that name with appropriate content be in the top level of the CPC  repo?

SaraJo:
> For the PR to move to stage 3, I think we should have the files which will end up living in the non-proposal sections of the CPC repo, for example should the SUPPORTER_PROGRAM.md or a file with that name with appropriate content be in the top level of the CPC repo?

It would need a few edits, but almost there. I can take a look at that this weekend.

mhdawson:
@SaraJo I hope you don't mind I pushed a few commits after talking with Robin/Jory today. You can discard them if they don't make sense.

It moves the SUPPORTER_PROGRAM.md file, which as you mentioned may need a few more edits as well as governance updates.

In terms of the governance updates we thought it was best to keep discussion of the 3rd board seat separate.  To this end the concept is something along these lines:

1) The CPC no longer believes an individual membership program makes sense as part of it's scope, instead it is replaced it with a supporter program that the CPC will manage.
2) Since there is no longer an individual membership program the 3rd board seat will never activate and therefore does not need to be mentioned in the CPC charter and is removed.
3) Since the Foundation bylaws mention a membership program which had previously been delegated by the board to the CPC to manage, and the CPC no longer has plans to start/manage an individual membership program, there will be a separate discussion on how to update the bylaws to re-purpose the 3rd seat.  Robin/Jory are going to work with a proposal on that. The new proposal may result in something being added back to the CPC charter at some point, but that does not block this landing in the mean time.

mhdawson:
I suggest we delete the required resources section from the file that will be at the CPC top level : https://github.com/openjs-foundation/cross-project-council/blob/a530c87ee508bd13ebc2248fec6943157241ad87/SUPPORTER_PROGRAM.md#required-resources, otherwise I think it's reasonable.

tobie (approved):

joesepi (approved):

mhdawson (approved):
LGTM once it is updated to reflect new proposal process directory structure

SaraJo:
> LGTM once it is updated to reflect new proposal process directory structure

On it, should I add it to "approved" or "incubating"?

mhdawson:
@SaraJo I think it should be "approved" as that is the current equivalent to stage3.  I also think its the right place since we have agreed on the program and it is moving forward :)

SaraJo:
@mhdawson Made the move to approved, mind taking a look?

mhdawson:
This has been open for quite some time, has approvals landing/

PR #696, opened 2020-12-15
Move Individual Membership to artifacts

joesepi:
in prep for new program. 

This will be replaced by the individual supporter program which is in its final stage in the proposal process


mhdawson (approved):
LGTM

ljharb (approved):

joesepi:
I'd like to request that we fast track this PR only so that we can be sure it is removed before any confusion happens with any other things that could be confusing. ;)

In order to fast track, we would need to CPC members to thumbs up this comment and also to have no objections to the fast tracking request.

https://github.com/openjs-foundation/cross-project-council/blob/master/GOVERNANCE.md#fast-tracking-prs

tobie (approved):
LGTM.

MylesBorins (approved):
LGTM

MylesBorins:
Do we want to add any commentary at the top as to why it is archived?

PR #717, opened 2021-02-16
Require consent rather than 2/3 vote for emeritus

### LINKED ISSUE #713: Criteria to move to Emeritus
PEP recently moved to Emeritus without any vote. Project Progression says there should be a vote. If we do not believe there should be a vote, then we should update Project Progression.

https://github.com/openjs-foundation/cross-project-council/blob/main/PROJECT_PROGRESSION.md#acceptance-criteria-3

Let's discuss. :)

eemeli:
Closes #713

### DIFF
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -213,7 +213,8 @@ Projects in this stage are not in active development. Their maintainers may infr
 
 #### Acceptance Criteria
 
-Projects may be granted Emeritus status via a 2/3 vote from the CPC and with approval from project ownership. In cases where there is a lack of project ownership, only a 2/3 vote from the CPC is required.
+Projects may be granted Emeritus status with the consent of the CPC and with approval from project ownership.
+In cases where there is a lack of project ownership, a 2/3 vote from the CPC is required.
 
 ### Incubating
 
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -214,7 +214,7 @@ Projects in this stage are not in active development. Their maintainers may infr
 #### Acceptance Criteria
 
 Projects may be granted Emeritus status through CPC consensus and with the approval of the maintainers of the project.
-In cases where there is a lack of project ownership, a 2/3 vote from the CPC is required.
+In cases where the project maintainers aren't responding after having been repeatedly contacted through appropriate channels about the CPC's intent to move the project to Emeritus stage, the CPC may proceed with the stage change without approval from the maintainers.
 
 ### Incubating
 
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -213,7 +213,7 @@ Projects in this stage are not in active development. Their maintainers may infr
 
 #### Acceptance Criteria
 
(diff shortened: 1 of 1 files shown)


sendilkumarn (approved):

PROJECT_PROGRESSION.md
```diff
@@ -213,7 +213,8 @@ Projects in this stage are not in active development. Their maintainers may infr
 
 #### Acceptance Criteria
 
-Projects may be granted Emeritus status via a 2/3 vote from the CPC and with approval from project ownership. In cases where there is a lack of project ownership, only a 2/3 vote from the CPC is required.
+Projects may be granted Emeritus status with the consent of the CPC and with approval from project ownership.
```
MylesBorins, line 216: Should this be "consensus" rather than consesnt?

codebytere (approved):

PROJECT_PROGRESSION.md
tobie: Suggesting the following wording to make your life easier. Feel free to ignore and word differently.

```suggestion
Projects may be granted Emeritus status through CPC consensus and with the approval of the maintainers of the project.
```

PROJECT_PROGRESSION.md
```diff
@@ -213,7 +213,8 @@ Projects in this stage are not in active development. Their maintainers may infr
...
+In cases where there is a lack of project ownership, a 2/3 vote from the CPC is required.
```
tobie, line 217: Clarifying the concept of "ownership" which isn't really defined anywhere:

```suggestion
In cases where the project maintainers aren't responding to the CPC for six (6) weeks after having been contacted through the appropriate channels about the CPC's intent to move the project to Emeritus stage, the CPC may proceed with the stage change without approval from the maintainers.
```

mhdawson (approved):
LGTM

mhdawson:
I think this changes our governance so we should get board approval.

PROJECT_PROGRESSION.md
eemeli: I think I agree with this in general, but it feels like a slightly bigger change to add this level of specificity. Also, is it intentional that you're proposing dropping the 2/3 vote requirement for the no-maintainers/owners case?

tobie: > I think I agree with this in general, but it feels like a slightly bigger change to add this level of specificity.

Would you prefer:

```suggestion
In cases where the project maintainers aren't responding after having been repeatedly contacted through appropriate channels about the CPC's intent to move the project to Emeritus stage, the CPC may proceed with the stage change without approval from the maintainers.
```

> Also, is it intentional that you're proposing dropping the 2/3 vote requirement for the no-maintainers/owners case?

Yes. I don't see the rationale for making a distinction between those two cases. If anything, Emeritus should be automatic for projects who are no longer maintained.

mhdawson:
@SaraJo and I have this on the agenda for the board meeting tomorrow.

mhdawson:
This was approved by the board in the last meeting. Either with/without the proposed changes from @tobie.

mhdawson:
In the CPC meeting there was consensus to add Tobie's comments and land based on the board approval. We can handle any further discussion/tweaks through a follow on.

Issue #745, opened 2021-04-20
Voting CPC Members Election


jorydotcom:
@eemeli pointed out that we missed the term deadline for Voting CPC members - we need to run an election!

eemeli:
Actually, we need to run [two elections](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/CPC-CHARTER.md#voting-members): One for the non-impact projects, and another for the regulars members.

Previously we've first done one, then the other. Is there a reason to change that this time, and run them simultaneously? For the non-impact project reps, [the process](https://github.com/openjs-foundation/cross-project-council/blob/main/GOVERNANCE.md#election-process-for-non-impact-voting-cpc-members) is pretty well documented and will take at least five weeks, but IIRC we don't have such explicit guidance about the regular member rep election.

jorydotcom:
I opened the issues to start the nomination period; next up I am going to open a PR to add some language as we discussed about what happens if we miss the nomination window.

PR #753, opened 2021-04-27
Update Voting Member Term info

jorydotcom:
Per #745 - we were looking to add prose that made it clear current Voting members are still valid if an election was not complete in time.


mhdawson (approved):
LGTM

eemeli (approved):

tobie (approved):
LGTM. Thanks for cleaning up leftovers from bootstrapping on your way!

PR #774, opened 2021-07-15
Country of origin

brettz9:
This is distinct from nationality, ethnicity, and race. For example, someone might be accused of harboring negative loyalties or political inclinations based on the political status of their country of origin, even if they have changed their nationality. It can differ from ethnicity or race in that the attack might be based more on politics rather than culture or racial identity.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -71,7 +71,7 @@ For more information, refer to the full
 
 ### Our Pledge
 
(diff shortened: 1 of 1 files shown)


MylesBorins:
This text is taken from the [upstream Contributor Covenant](https://github.com/EthicalSource/contributor_covenant/blob/release/content/version/2/0/code_of_conduct.md)

In the past we have worked upstream for changes like this rather than floating a patch.

ljharb:
→ https://github.com/EthicalSource/contributor_covenant/pull/967

josepachecojp0732 (commented):
!

PR #775, opened 2021-07-16
Add ESLint reporting email

brettz9:


nzakas (approved):

tobie (commented):
I'm uneasy with the lack of consistency we have displaying the information for the different projects which are delegating CoC violation handling to the Foundation's Code of Conduct Panel. It seems some add the CoCP's email as their reporting email while some don't.

Could we decide what makes the most sense for a someone wanting to report a violation and standardize accordingly?

nzakas:
Yes please. I was under the belief that there was one place to report CoC violations at the foundation and was surprised to find these listed separately. My opinion is that for most projects, it makes sense for the foundation to receive any initial complaint to investigate, as not all projects have the bandwidth or experience to deal with CoC violations themselves.

jorydotcom:
Projects _may_ delegate their CoC violation reporting to the CPC; many projects have their own reporting methods while others don't have the bandwidth and choose to delegate to the Foundation (which is a benefit we offer to projects). The standardization happens on the Code of Conduct itself, and on how the CPC adjudicates reports. This allows the projects to independently manage their communities with support from the Foundation and broader community when asked. Does that answer your question @tobie? @nzakas that means ESLint is welcome to delegate reporting to the CPC as Grunt and some other projects do; however if there is bandwidth and interest in managing it within your technical community you can do that (and still leverage us later, if you need a helping hand)

joesepi (approved):

joesepi:
@jorydotcom reading your comment above, it sounds like some projects delegate reporting to the foundation, such as Grunt, but do not reflect that in this document. Is that correct? (Grunt being the example, does not have a reporting email here.)

It would seem that every project should have an email listed here. These comments are beyond this PR, so I've just opened an issue: #781

nzakas:
@jorydotcom thanks for the explanation. ESLint would definitely like to delegate that to the CPC, so is this PR correct now? If so, I'd also suggest a reformatting of the page outside of this PR to have the CPC email front and center, and then only list projects for which the CPC is not handling CoC complaints.

tobie:
We're in agreement about the mechanics, @jorydotcom.

My issue is that out of those projects that delegate to the CoCP, some list the CoCP's email, and some don't. This is potentially confusing and could be improved. Following up on the issue @joesepi opened.

jorydotcom (approved):

jorydotcom:
@nzakas yep! and great suggestion. It is hard to read.

@tobie - good to hear; glad for the comms followup on a different issue. Thanks!

joesepi:
I'm going to merge this and I believe @tobie has a PR open to address the other comments.

Issue #824, opened 2022-01-18
Add missing dates to dates and reminders doc


joesepi:
Some dates are missing here: https://github.com/openjs-foundation/cross-project-council/blob/main/Dates-and-Reminders.md

joesepi:
Current data in dates and reminders doc:
* CPC Chair election
* Director 1 election - term ends February 28
* Director 2 election - term ends August 31
* CPC Impact voting member selection - terms end March 31
* CPC Non-impact voting member election

---

- CPC Election cycle
2021, Joe was “elected” in August. [CPC Chairperson election 2021](https://github.com/openjs-foundation/cross-project-council/issues/772)
Director2 term ends Aug 31
Do we want to align with that or stagger?


- CPC Non-impact voting members cycle
Tobie and Eemeli were elected  [on May 14, 2021](https://github.com/openjs-foundation/cross-project-council/issues/751#issuecomment-841501713)

PR #839, opened 2022-03-15
Update dates for terms

joesepi:
In today's CPC meeting, we discussed aligning a few term ends to coincide with the OpenJS World event. The thinking is that these specific roles are good roles to recruit and promote at the event.

I've also picked a somewhat arbitrary date for the CPC chair role term.

See some more notes in the following issue:
closes: #824


Trott (approved):

eemeli (approved):

mhdawson (approved):
LGTM

PR #877, opened 2022-04-27
bump linkinator package

WaleedAshraf:


.github/workflows/link-checker.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: JustinBeckwith/linkinator-action@v1
+      - uses: JustinBeckwith/linkinator-action@v1.10.0
```
ljharb, line 16: i'm confused, "v1" should always mean "latest v1". Why would we need to explicitly update it?

https://github.com/JustinBeckwith/linkinator-action/tree/v1 eg already points to 1.10.0

ljharb: Seems like transient errors, since the links that 404 do in fact work. I reran that failed job and it passed.

PR #944, opened 2022-10-07
Fix nodejs link

AugustinMauroy:
# Fix nodejs link
nodejs.org to nodejs.dev


WaleedAshraf (commented):
nodejs.dev is still under development. We have not officially migrated to the new site, yet.

PR #964, opened 2022-11-11
Add Regular Member voting and clarify At large term

joesepi:
It seems that the regular member voting election isn't in this list. 🤔 

Also, I believe we discussed removing the "non-Impact" term in favor of At-Large since we removed the Growth designation a while ago.

Signed-off-by: Joe Sepi <sepi@joesepi.com>


eemeli (approved):

mcollina (approved):
lgtm

ljharb (approved):

mcollina (approved):
lgt

ovflowd (commented):
:shipit:

tobie (commented):
@joesepi wrote:
> I believe we discussed removing the "non-Impact" term in favor of At-Large since we removed the Growth designation a while ago.

My understanding is we're doing some back and forth on that depending on whether we're thinking about being inclusive to incubating projects or clear that growth status is no longer a thing.

tobie (approved):

dylans (approved):

ovflowd (approved):

mhdawson (approved):
LGTM

PR #1007, opened 2023-02-08
Minimise diff between our CoC & upstream by adjusting titles & newlines

eemeli:
As identified in far too many discussions, our upstream CoC is [not stable](https://github.com/EthicalSource/contributor_covenant/commits/release/content/version/2/0/code_of_conduct.md). This PR does not apply any changes to the textual content of our CoC, but adjusts the line wrapping and title levels to minimise the difference between our text and the [latest upstream text](https://raw.githubusercontent.com/EthicalSource/contributor_covenant/192bcf8016ae3733e46ff7d1285f2f57152cc195/content/version/2/1/code_of_conduct.md), v2.1.

After this is merged, it becomes somewhat easier to consider each of the four remaining parts of the diff separately:

### 1. Embedding the doc to within our `CODE_OF_CONDUCT.md`

Skipping here all the text before the `---` separator.

```diff
--- /tmp/upstream-coc.md	2023-02-08 14:22:10
+++ /tmp/openjsf-coc.md	2023-02-08 14:41:36
@@ -1,17 +1,12 @@
-+++
-version = "2.1"
-aliases = ["/version/2/1"]
-+++
-
-# Contributor Covenant Code of Conduct
+_Contributor Covenant Code of Conduct v2.0, from [commit 8001bf8](https://github.com/EthicalSource/contributor_covenant/blob/8001bf8c6e7cd2606657e2816710770d8a79b7dc/content/version/2/0/code_of_conduct.md)_

 ## Our Pledge
```

### 2. Adding our contact methods

```diff
--- /tmp/upstream-coc.md	2023-02-08 14:22:10
+++ /tmp/openjsf-coc.md	2023-02-08 14:41:36
@@ -65,7 +60,9 @@

 Instances of abusive, harassing, or otherwise unacceptable behavior may be
 reported to the community leaders responsible for enforcement at
-[INSERT CONTACT METHOD].
+the email addresses listed above in the
+[Reporting](#report-an-issue-in-a-project) and
+[Escalation](#escalate-an-issue) sections.
 All complaints will be reviewed and investigated promptly and fairly.

 All community leaders are obligated to respect the privacy and security of the
```

### 3. Differences between the v2.0 and v2.1 versions of the upstream

The terms "caste" and "color" were added in v2.1.

```diff
--- /tmp/upstream-coc.md	2023-02-08 14:22:10
+++ /tmp/openjsf-coc.md	2023-02-08 14:41:36
@@ -10,17 +5,12 @@
 We as members, contributors, and leaders pledge to make participation in our
 community a harassment-free experience for everyone, regardless of age, body
 size, visible or invisible disability, ethnicity, sex characteristics, gender
 identity and expression, level of experience, education, socio-economic status,
-nationality, personal appearance, race, caste, color, religion, or sexual
+nationality, personal appearance, race, religion, or sexual
 identity and orientation.

 We pledge to act and interact in ways that contribute to an open, welcoming,
@@ -120,13 +117,13 @@
 ## Attribution

 This Code of Conduct is adapted from the [Contributor Covenant][homepage],
-version 2.1, available at
-[https://www.contributor-covenant.org/version/2/1/code_of_conduct.html][v2.1].
+version 2.0, available at
+[https://www.contributor-covenant.org/version/2/0/code_of_conduct.html][v2.0].

 Community Impact Guidelines were inspired by
 [Mozilla's code of conduct enforcement ladder][Mozilla CoC].
@@ -131,7 +128,7 @@
 [https://www.contributor-covenant.org/translations][translations].

 [homepage]: https://www.contributor-covenant.org
-[v2.1]: https://www.contributor-covenant.org/version/2/1/code_of_conduct.html
+[v2.0]: https://www.contributor-covenant.org/version/2/0/code_of_conduct.html
 [Mozilla CoC]: https://github.com/mozilla/diversity
 [FAQ]: https://www.contributor-covenant.org/faq
 [translations]: https://www.contributor-covenant.org/translations
```

### 4. Retaining the "project" specifier for perma-bans

This was [dropped from the upstream](https://github.com/EthicalSource/contributor_covenant/commit/a63f697fae93de5abe6354af2c4afa78c5804298) after our copy was taken as an apparently missed part of the v1.4 -> v2.0 updates.

```diff
--- /tmp/upstream-coc.md	2023-02-08 14:22:10
+++ /tmp/openjsf-coc.md	2023-02-08 14:41:36
@@ -115,13 +112,13 @@
 individual, or aggression toward or disparagement of classes of individuals.

 **Consequence**: A permanent ban from any sort of public interaction within the
-community.
+project community.

 ## Attribution
```

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -43,81 +43,137 @@ For more information, refer to the full
 
 ---
 
-## Contributor Covenant Code of Conduct v2.0
+_Contributor Covenant Code of Conduct v2.0, from [commit 8001bf8](https://github.com/EthicalSource/contributor_covenant/blob/8001bf8c6e7cd2606657e2816710770d8a79b7dc/content/version/2/0/code_of_conduct.md)_
 
-### Our Pledge
+## Our Pledge
 
-We as members, contributors, and leaders pledge to make participation in our community a harassment-free experience for everyone, regardless of age, body size, visible or invisible disability, ethnicity, sex characteristics, gender identity and expression, level of experience, education, socio-economic status, nationality, personal appearance, race, religion, or sexual identity and orientation.
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual
+identity and orientation.
 
-We pledge to act and interact in ways that contribute to an open, welcoming, diverse, inclusive, and healthy community.
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
 
-### Our Standards
+## Our Standards
 
-Examples of behavior that contributes to a positive environment for our community include:
+Examples of behavior that contributes to a positive environment for our
+community include:
 
 * Demonstrating empathy and kindness toward other people
 * Being respectful of differing opinions, viewpoints, and experiences
 * Giving and gracefully accepting constructive feedback
-* Accepting responsibility and apologizing to those affected by our mistakes, and learning from the experience
-* Focusing on what is best not just for us as individuals, but for the overall community
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the overall
+  community
 
 Examples of unacceptable behavior include:
 
-* The use of sexualized language or imagery, and sexual attention or advances of any kind
+* The use of sexualized language or imagery, and sexual attention or advances of
+  any kind
 * Trolling, insulting or derogatory comments, and personal or political attacks
 * Public or private harassment
-* Publishing others' private information, such as a physical or email address, without their explicit permission
-* Other conduct which could reasonably be considered inappropriate in a professional setting
+* Publishing others' private information, such as a physical or email address,
+  without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
 
-### Enforcement Responsibilities
+## Enforcement Responsibilities
 
-Community leaders are responsible for clarifying and enforcing our standards of acceptable behavior and will take appropriate and fair corrective action in response to any behavior that they deem inappropriate, threatening, offensive, or harmful.
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
 
-Community leaders have the right and responsibility to remove, edit, or reject comments, commits, code, wiki edits, issues, and other contributions that are not aligned to this Code of Conduct, and will communicate reasons for moderation decisions when appropriate.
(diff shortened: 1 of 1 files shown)


eemeli:
Requesting fast-tracking for this, as its a purely stylistic change which makes comparisons to the upstream much easier.

ljharb (approved):
The soft wrapping was an intentional decision - hard wrapping is horrific for editing and diff churn - but i agree that creating an excessive diff from upstream isn't worth the benefit.

CODE_OF_CONDUCT.md
```diff
@@ -43,81 +43,137 @@ For more information, refer to the full
 
 ---
 
-## Contributor Covenant Code of Conduct v2.0
+_Contributor Covenant Code of Conduct v2.0, from [commit 8001bf8](https://github.com/EthicalSource/contributor_covenant/blob/8001bf8c6e7cd2606657e2816710770d8a79b7dc/content/version/2/0/code_of_conduct.md)_
```
anonrig, line 46: Do we want this change in the CODE_OF_CONDUCT? Explicitly adding a version label, other than the upstream changes will create diffs, and the previous pull request needs to remove this particular change. Due to this, I don't see any good reasoning for making this change.

anonrig (changes requested):

CODE_OF_CONDUCT.md
ljharb: This seems like maybe something that the patch script could add implicitly? then it wouldn't be shown as a diff in the patch file but would still help track this info.

tobie (approved):
<3 <3 <3

ovflowd (approved):

CODE_OF_CONDUCT.md
eemeli: Much of the confusion in our CoC discussions has arisen from not having clarity on what changes in our CoC compared to the upstream are modifications we've made, and what has been modified in the upstream. Our current text is an essentially unmodified copy of the upstream at the specified commit, but the upstream "v2.0" has since been modified. If/as we start to actively patch our CoC, I would think that we continue to need an explicit indicator for where our text is coming from.

If we do start to use a script to apply such changes, this text will need to reflect that, possibly directing interested partied to a more detailed explanation for where to find the sources.

eemeli:
> @ljharb:
> The soft wrapping was an intentional decision - hard wrapping is horrific for editing and diff churn - but i agree that creating an excessive diff from upstream isn't worth the benefit.

Actually, we never changed the wrapping from how it was when we copied it. It's the upstream that has subsequently applied hard wrapping to the text. This change aligns our wrapping with the current state of the upstream v2.1 text, effectively backporting stylistic updates to our copy.

CODE_OF_CONDUCT.md
anonrig: > If we do start to use a script to apply such changes, this text will need to reflect that

This requires duplication of updating the CoC version, and I believe it might be a better way to use the update script to append the results for the versioning of it automatically.

ljharb:
@eemeli ah, i'd remembered it wrong. then it's just a really unfortunate upstream decision.

bnb:
I believe we agreed to ship this today because it's supposed to minimize the diff prior to landing #972

joesepi (approved):

anonrig (approved):

PR #1021, opened 2023-03-02
meta: update links

VoltrexKeyva:
- Update the current year fund link to point to the 2023 fund.
- List all previous funds.

### DIFF
--- a/project-resources/MEMBER_TRAVEL_FUND.md
+++ b/project-resources/MEMBER_TRAVEL_FUND.md
@@ -5,11 +5,18 @@
 To establish and administer a fund for members of the OpenJS Foundation to travel
 and spread knowledge about and support the Javascript ecosystem and the Foundation.
 
-### [Current Year Fund](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/2022.md)
+### [Current Year Fund](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/2023.md)
 
-### [2021 funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2021.md)
+<details>
+  <summary>Previous funds</summary>
 
-### [2019 funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2019.md)
+  * [2022](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2022.md)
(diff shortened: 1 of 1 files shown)


Trott:
> * The vis.gl incubating project's page no longer sends secure HTTPS responses, this causes the Linkinator GitHub action to fail, so it's project link protocol should be changed to HTTP.

HTTPS service seems to be restored for that website now.

VoltrexKeyva:
@Trott reverted the change to the link protocol, PTAL.

project-resources/MEMBER_TRAVEL_FUND.md
```diff
@@ -5,11 +5,11 @@
...
 and spread knowledge about and support the Javascript ecosystem and the Foundation.
 
-### [Current Year Fund](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/2022.md)
+### [Current Year Fund](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/2023.md)
 
-### [2021 funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2021.md)
+### [2022 Funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2022.md)
 
-### [2019 funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2019.md)
+### [2021 funds](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/TRAVEL_FUND/Achived_Travel_Fund_Records/2021.md)
```
ljharb, line 12: did we not have a 2020 fund?

seems like we should list all of them, just just the prior 2 years?

VoltrexKeyva: There doesn't seem to be any sign of a 2020 fund, so I'm guessing there wasn't one.

Also I'm not sure if we should list all of them, I think we should keep listing the most relevant ones to date because there's a high chance most people don't look at the old ones, and that this list will keep getting bigger on every new fund entry.

ljharb: Why bother to list the two previous years then?

If 2021 is relevant than so is 2019 imo. If the list gets too big, that's what `<details>` is for.

VoltrexKeyva: Hmm I suppose you're right, though should we add a `<details>` tag with a `<summary>` of "Previous funds" now? So that we don't have to add one later.

ljharb: sure, sounds fine

VoltrexKeyva: Done.

ljharb (approved):

ovflowd (approved):
LGTM!

eemeli (approved):

PR #1050, opened 2023-04-04
doc: revert changes on the preamble

anonrig:
Upon request from @eemeli, this pull request reverts changes on the preamble file so that the chain of patching script is not broken.

### DIFF
--- a/conduct/artifacts/CODE_OF_CONDUCT_PREAMBLE.md
+++ b/conduct/artifacts/CODE_OF_CONDUCT_PREAMBLE.md
@@ -1,6 +1,6 @@
 # Code of Conduct
 
-The OpenJS Foundation and its member projects use a modified version of the Contributor Covenant as their code of conduct, and the translations from the original script will not reflect all modifications. The full text is included below in English, and [the original translations](https://www.contributor-covenant.org/translations) are available on the Contributor Covenant website.
+The OpenJS Foundation and its member projects use [Contributor Covenant v2.0](https://contributor-covenant.org/version/2/0/code_of_conduct) as their code of conduct. The full text is included below in English, and [translations](https://www.contributor-covenant.org/translations) are available on the Contributor Covenant website.
 
 ## Commitment
 


ljharb (approved):

joesepi (approved):

eemeli:
Is there a reason why the same changes to the preamble file that I included in #1047 are not suitable, and they need a separate PR?

anonrig:
> Is there a reason why the same changes to the preamble file that I included in #1047 are not suitable, and they need a separate PR?

Because that commit also included the changes on the CODE_OF_CONDUCT file. After landing this patching script, on every commit to `main` branch, the patching script will run its workflow and open a pull request according to those changes. This was the intention of the patching script workflow that was implemented, and I personally think we should stick to it until we decide to remove it.

eemeli:
So now if we get an upstream CoC change that we _don't_ want to adapt, the process is something like this?

1. Have the bot create a PR for it
2. Create a separate PR modifying the preamble or patch
3. Approve & merge the second PR
4. Have the bot update the first PR
5. Close the first PR as it's now an empty changeset

This seems like a lot of hassle that isn't making our life easier than just being notified of an upstream change.

anonrig:
If we don't want to adapt a certain change, than we shouldn't merge it to main branch in the first place. In this particular example, the change was proposed, accepted and later merged to the main, but because we didn't give access to the bot to push to main directly, it opened a PR, causing an illusion for needing to review again, but it is already a change that is reviewed and merged into the main branch.

eemeli:
The script should be looking at the preamble and patch in its PR branch rather than `main`.

anonrig:
> The script should be looking at the preamble and patch in its PR branch rather than `main`.

Sorry, but I don't follow. Can you elaborate on this?

eemeli:
> > The script should be looking at the preamble and patch in its PR branch rather than `main`.
> 
> Sorry, but I don't follow. Can you elaborate on this?

Normally, when the script is run, it compares the state of the `main` branch to the upstream. If, when it's run, the `create-or-update-pull-request-action` branch exists, it means that a prior change has not yet been merged in yet. In that situation, the script should base its actions on the PR branch, and not `main`.

Consider what would happen if #1047 were merged now. It's currently updating both the CoC and the preamble, but differently, because it's updated the CoC based on `main` after I added a commit to the PR also updating the preamble.

joesepi:
@eemeli - Changes should not be made on PRs opened by the bot from this workflow. That will put things out of sync.

Additionally, if there is an open PR from the bot and another change happens to the underlying files (outside of the existing PR), then the patching script workflow will update the existing open pull request

@anonrig and I just opened a PR with some documentation on this effort https://github.com/openjs-foundation/cross-project-council/pull/1051

ljharb:
To be even clearer, the bot's PR *should never be anything but approved*, because all the related changes were already approved and merged. Once the approach is validated, this bot-created PR will be obviated by making the bot autocommit and push, avoiding this confusion.

bnb (approved):

ovflowd (approved):

PR #1070, opened 2023-05-10
fix(readme): add `alt` to `<img />`

AugustinMauroy:


README.md
```diff
@@ -1,7 +1,7 @@
 <a href="https://openjsf.org">
   <picture>
     <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color-dark_background.svg" />
-    <img src="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color.svg" width="200" />
+    <img src="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color.svg" width="200" alt="Openjs Foundation Logo" />
```
ljharb, line 4:
```suggestion
    <img src="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color.svg" width="200" alt="OpenJS Foundation Logo" />
```

ljharb, line 4:
```suggestion
    <img src="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color.svg" width="200" alt="OpenJS Foundation Logo" />
```

ljharb (approved):

README.md
AugustinMauroy: In trademark the name of foundation is"Openjs foundation" so I'm not fan of "openJS".
what do you think ?

Trott: I won't block this, but I'd prefer `alt=""` as the logo is decorative and does not provide information.

Trott: > In trademark the name of foundation is"Openjs foundation" so I'm not fan of "openJS".
> what do you think ?

What makes you say it's "Openjs foundation" rather than "OpenJS Foundation"?

Trott: > I won't block this, but I'd prefer `alt=""` as the logo is decorative and does not provide information.

Oh, actually, since the logo is a link back to the home page, it should be `alt="Home"` or something like that. The `alt` text should describe the information provided by the image (in this case: none) and the function of following the link.

ljharb: The name of the foundation is "OpenJS Foundation", so the "js" should never be lowercased.

I'm pretty sure the alt on the image should only describe the image; the link itself describes the link. An empty alt is fine too.

Trott:
```suggestion
    <img src="https://raw.githubusercontent.com/openjs-foundation/artwork/main/openjs_foundation/openjs_foundation-logo-horizontal-color.svg" width="200" alt="Openjs Foundation website" />
```

@ljharb Would you concur with this change based on the fact that the image is a link to the foundation website? If not, let's you and I discuss offline so Augustin isn't getting different messages from both of us.

Trott (changes requested):
Proposed alt text isn't appropriate in my view. Suggestion for appropriate alt text provided.

README.md
ljharb: no, the image isn't a link, the link is the link. happy to discuss offline.

Trott: > I'm pretty sure the alt on the image should only describe the image; the link itself describes the link. An empty alt is fine too.

[WebAIM says otherwise](https://webaim.org/techniques/alttext/): 

> Since no adjacent text within the link describes the link's function, it must be conveyed within the image alt attribute.

I'd be OK with "OpenJS Foundation" or "OpenJS Foundation website", but the word "logo" should not be included IMO. A screen reader, for example, would say something like "Link: OpenJS Foundation logo" which is not what we want.

Trott: Reading more of WebAIM, it seems like "website" should be left off. I'll update my suggestion.

ljharb: I like eliding “logo”

Trott (approved):

ljharb (approved):

Trott:
This is such a contained, low-risk, and easy-to-revert change that I think we can land it straightaway.

PR #1157, opened 2023-08-29
Delete OpenJSF Marketing Services - Dec. 2019.pdf

kyliewd:
Deleting outdated file.


ljharb:
While we could certainly delete it, it's got the date in the filename so nobody will be confused, and it seems nice to have records for archival purposes.

tobie (approved):

tobie:
> it seems nice to have records for archival purposes.

Isn't that what versioning is for? (This is not a rhetorical question, discovery is part of archival and versioning kind of sucks at that.)

ljharb:
Sure, if PDF diffs worked well on github then keeping the same file and changing it would be ideal, but forcing people to git spelunk to find old copies seems unfortunate.

joesepi:
I think if we were to keep the file, we would want to make it extra clear that it is outdated and no longer applicable. Like nest it in some directories: Project Resources > Archive > No longer used > file

But I dont think this document is useful once its been revamped. If we wanted to go back in time to see what the policy was, we could look back in the repo when it did exist. But I guess that is the git spelunking referred to above.

I'm fine with either approach.

PR #1159, opened 2023-09-01
Rename and update marketing slide deck

tobie:
This makes it easier to look for past versions of the marketing slide deck while only keeping the most recent copy directly available in the repository.

Closes #1157.
Closes #1158.


tobie:
My assumption is we probably shouldn't squash this PR, but I might be wrong.

joesepi (approved):

ljharb (approved):
indeed, if it's squashed the git history likely won't work

Issue #1170, opened 2023-09-12
Evaluate BlueOak License


bensternthal:
This issue tracks the legal evaluation of the BlueOak license https://blueoakcouncil.org/license/1.0.0 
```[tasklist]
### Tasks
- [x] Update https://github.com/openjs-foundation/cross-project-council/edit/main/governance/IP_POLICY_GUIDANCE.md
- [x] Update https://ip-policy.openjsf.org/
```

tobie:
What's the context for this? It's not an OSI-approved license and has very low usage.

MylesBorins:
transitive dependencies in npm, and a variety of other JS projects, were relicensed (legally) as Blue Oak.

https://github.com/nodejs/node/issues/49625

It is going to be very difficult to untangle this, much easier if OpenJS take a supportive position on the license

tobie:
I'm going to repeat what I said in the Node issue itself: the first step would be to get the license approved by OSI. (The process is here: https://opensource.org/licenses/review-process/.)

OpenJSF shouldn't substitute itself to OSI's process. That would be setting a really bad precedent.

MylesBorins:
> OpenJSF shouldn't substitute itself to OSI's process. That would be setting a really bad precedent.

If the OpenJS position is that only OSI approved licenses are allowed, that would allow for clarity. What I'd like to see is clarity on what is / is not allowed

MylesBorins:
FWIW I did a quick evaluation of other OpenJS Impact projects

Every single one of the projects rely on `glob`, the package that is dependent on packages licensed as blue-oak.

Appium, Node.js, and Electron all depend on version of glob that include a dependency with a blue oak license.

I haven't looked into at large or incubation projects but I'm sure this is going to be pervasive within OpenJS as well as the ecosystem at large (a recent example from [angular](https://github.com/angular/angular-cli/issues/25387).

So I do want to point out that if OpenJS wants to take a position that it does not allow the blue oak license, specifically as it isn't OSI approved, we are going to have a fairly extensive ecosystem problem due to `glob`.

mhdawson:
@bensternthal how long do we think it might take to get a legal opinion from our OpenJS lawyers on BlueOak and the risk given the current situation.
 
[...]

PR #1245, opened 2024-01-31
Update IP_POLICY_GUIDANCE.md

bensternthal:
- Adds blue oak license to IP policy guidance
- Fixes reference to a PM that is no longer with LF
- See #1170

### DIFF
--- a/governance/IP_POLICY_GUIDANCE.md
+++ b/governance/IP_POLICY_GUIDANCE.md
@@ -19,6 +19,7 @@ There are four key requirements for all OpenJS Foundation projects:
 
 Code may be licensed under:
   * [Apache, Version 2.0](http://www.apache.org/licenses/LICENSE-2.0)
+  * [Blue Oak Model License 1.0.0](https://opensource.org/license/blue-oak-model-license/)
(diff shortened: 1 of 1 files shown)


mhdawson (approved):
LGTM

ljharb (approved):

tobie (changes requested):
Let's wait for the IP policy at https://ip-policy.openjsf.org to be updated. _(I'm also unsure if the decision is to allow projects to use that license or just dependencies of projects to do so.)_

To be clear, I'm just blocking so we don't end-up with a guidance document that's ahead of the policy itself, not because of the proposed changes themselves.

bensternthal:
@tobie good call, I am working to get the IP policy reviewed and updated. Once that is live I will merge this PR.

tobie (approved):
Both concerned addressed. Thank you!!

UlisesGascon (approved):

joesepi (approved):

mcollina (approved):
lgtm

mcollina (approved):
lgtm

Issue #1241, opened 2024-01-24
Refine CPC Definition Of Consensus 


bensternthal:
Per comment on the travel fund pr in response to the following proposed workflow:

              

I looked at what we have in the Node.js project and it has the link to the wiki page but nothing else. My key point in the in the recent working group meeting is that the "consensus" must represent that of the CPC, not just the subset of CPC members who could make a particular meeting. That requires that people not in attendance have the opportunity to object before the decision is considered final. I think we might want to update the description in https://github.com/openjs-foundation/cross-project-council/blob/main/governance/GOVERNANCE.md#consensus-seeking-process to reflect that as well. For example, if two people show up at a CPC meeting they are not able to make a decision on behalf of all of the CPC.  Further even if 51% of the people show up at a meeting the other 49% of the members should still have an opportunity to object before the decision is considered final.

_Originally posted by @mhdawson in https://github.com/openjs-foundation/cross-project-council/pull/1230#discussion_r1456605087_

tobie:
I'd remove the travel-fund label and push it as a broader governance issue. This is how the CPC operates anyway, so this is more about making paper governance better spell-out actual governance.

tobie:
Following-up on our conversation in the CPC working session today, I'd suggest [adapting the PR merge request process](https://github.com/openjs-foundation/cross-project-council/blob/main/governance/GOVERNANCE.md#merging-prs-into-this-repository) for  travel fund requests. The PR merge process says the following:

> Pull requests that do not change the charter or governance of the CPC can be merged into this repository provided the following conditions have been met:
>
> - There are no outstanding objections
> - There are two approvals by CPC members
> - The PR has been open for at least 72 hours
>
> Pull requests that change governance of the CPC (including approving changes to Working Groups) must meet the following conditions in addition to the ones listed for regular PRs.
> 
> - The PR has been open for at least 14 days OR consensus is reached in a meeting with quorum of voting members.

I'd suggesting the following for approving travel fund requests:

> Funding requests that have a rubric score of _x_ or higher are approved provided the following conditions have been met:
>
> - There are no outstanding objections
> - There are two approvals by CPC members
> - The email has been sent for at least 72 hours
>
> Funding requests that have a rubric score lower than _x_ must meet the following conditions in addition to the ones listed above:
> 
> - The email has been open for at least 14 days OR consensus is reached in a meeting with quorum of voting members.

ljharb:
[...]

PR #1281, opened 2024-03-28
doc: improve doc for lazy consensus process

mhdawson:
Refs: https://github.com/openjs-foundation/cross-project-council/issues/1241

### DIFF
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -208,9 +208,7 @@ Changes to this document require CPC consensus and approval from the OpenJS Foun
 
 [Foundation mission and vision statements]:  https://openjsf.org/about/
 [Foundation bylaws]:  https://bylaws.openjsf.org
-[Voting]:  #section-10.-voting
-[Elections]: #section-7.-elections
-[AS community]: https://community.apache.org/committers/decisionMaking.html#lazy-consensus
+[Elections]: #section-7-elections
 [Condorcet]: http://en.wikipedia.org/wiki/Condorcet_method
 [Single Transferable Vote]: http://en.wikipedia.org/wiki/Single_transferable_vote
 [OpenJS Foundation bylaws]: https://bylaws.openjsf.org/


governance/GOVERNANCE.md
```diff
@@ -57,17 +57,21 @@ To improve organizational transparency and collaboration between the CPC and Boa
...
-If an agenda item cannot reach a consensus, a Team member can call for
-the item to be decided by a vote or to table the issue to the next
-meeting. In both cases the decision must be seconded by a majority of the Team
-or else the discussion will continue. Simple majority wins. Only Active
-Members participate in a vote.
+The Team follows a [Lazy Consensus Seeking][] decision-making model.
+
+Lazy consensus is achieved by making a proposal in a forum that will reach all
+members of the group who should have the opportunity to participate in reaching
+consensus and waiting for an appropriate amount of time for anyone to object.
```
ljharb, line 64: How is “appropriate” determined?

ljharb (approved):

governance/GOVERNANCE.md
marco-ippolito: I think we should define a number like X weeks, and if it's too much or too little we can update it

mcollina: I prefer to keep it loose here, and mention something like "after a CPC public meeting".

mhdawson: I had thought about it and different times may apply depending on the group/topic which is why I left it out I think specific times can be defined for specific cases, for example approving a new project might have a longer time than something else.

ljharb: I'm totally fine to keep it loose in the actual document, to be clear, which is why i stamped it - i was trying to get an understanding of what others' intentions/expectations here are.

governance/GOVERNANCE.md
```diff
@@ -57,17 +57,21 @@ To improve organizational transparency and collaboration between the CPC and Boa
...
+It is not necessary to get explicit approval to proceed from all members,
+silence indicates consent. However, objections must be resolved in order
+to achieve consensus.
+
+If a proposal cannot reach a consensus, a group member can call for
+the proposal to be decided by a vote. This should only be done as a last resort
+when consensus cannot be reached after a reasonable period of discussion.
+For all votes, a simple majority of all group voting members for, or against,
+the issue wins. A group voting member may choose to participate in any vote
+through abstention.
```
tobie, line 74: > "A group voting member may choose to participate in any vote
through abstention."

I'm not exactly sure what this means.

governance/GOVERNANCE.md
tobie: +1 to keeping the definition non-specific here. We could indicate that the timing is context-specific, like so:

```suggestion
consensus and waiting for an appropriate amount of time given the nature of the decision for anyone to object.
```

tobie (commented):
Should we modify [the charter](https://github.com/openjs-foundation/cross-project-council/blob/main/CPC-CHARTER.md#section-9-decision-making) accordingly?

tobie (approved):

governance/GOVERNANCE.md
ljharb: To me it means that you can participate without saying yes or no, by abstaining.

tobie (changes requested):
Thanks a lot for getting this started, @mhdawson.

Giving this another thought, I think we shouldn't duplicate the charter's content in the `governance.md` at all, even as a first step. `governance.md` should only reference the decision-making and voting policies of the charter.

I suggest we:

1. change this PR so that `governance.md` only references the decision-making and voting policies of the charter, and
2. bring your suggested changes directly into the charter itself, either in this PR or in a new, separate PR (whatever you feel as more convenient).

governance/GOVERNANCE.md
tobie: So that's the language that's also in the charter. It's either too explicit or not enough. For example, 

- is the vote of present members or all members?
- is the vote async or not?
- if the vote is of members present, do we need a quorum to have a vote?
- do votes need to be announced in advance?
- is simple majority "more than 50%" or "the answer with a bigger score"?
- if majority is "more than 50%" are abstentions counted?
- if so, what prevents reframing the question as its negative to flip the result of the vote, e.g. 4 people vote for prop `x`, 3 against, and 3 abstain. The vote fails with only 40% of the votes. Now organize a vote for `!x`. You'll get the 4 people that were for `x` vote against `!x`, the 3 that were against `x` now vote for `!x`, and the same 3 as before will abstain. So now we have only 30% for `!x` which fails, an do so `x` passes.
- can anyone request a vote at anytime? If so, that goes against the idea of being consensus-driven. If not, the language needs to be mores specific about when it is OK to ask for a vote.

I don't want us to spend too much time discussing voting as the case never really showed up, so my suggestion would be to acknowledge that voting should be a last resort solution, costly in social capital, painful to organize, hard to weaponize, and favoring the status quo.

I'd suggest the following language:

> Voting should only be done as a last resort. If a proposal to modify existing policies and processes (or add new ones) cannot reach consensus after a reasonable period of discussion, a voting member can call for objections to be overruled through an asynchronous vote. If this call is seconded by another voting member, an asynchronous vote must be organized. For the objections to be overruled (and thus the proposal to pass), at least 2/3 of all voting members must vote in favor. This is designed to make it difficult to change existing policies and processes (or to add new ones) without consensus, but prevent someone from completely blocking progress through objections.

WDYT?

tobie:
Opened a PR against this PR here: https://github.com/mhdawson/bootstrap/pull/1

marco-ippolito (approved):

governance/GOVERNANCE.md
UlisesGascon: > WDYT?

I think the suggestion is a great addition.

mhdawson:
@tobie landed your PR.

tobie (approved):

UlisesGascon (approved):

wesleytodd (approved):

joesepi (approved):

mhdawson:
We discussed in the CPC working meeting today. The main point of discussion was the requirement for 2/3 which is different than what was previously documented. None of the 9 people in the meeting were concerned with that. @openjs-foundation/cpc if you agree please add your +1 to the PR or make suggestions. We agreed to make one more checkin with the CPC before landing the PR based on no objections/approvals.

tobie:
Heads-up that while @mhdawson's original PR didn't touch the charter this final version does, and so this needs Board approval.

anonrig (approved):

governance/GOVERNANCE.md
PaulaPaul: I also think this is a great addition/clarification of the consensus process for decision making. 
It's a nit and not needed to land these changes, but depending on who is looking at the governance doc and decision making processes, it could help to explicitly state that these processes are applicable to anything in scope for the CPC (defined in the README.md). Nice work!

tobie: Technically, the scope of the CPC is defined in the charter itself.

tobie:
We just got board approval in today's board meeting.

PR #1329, opened 2024-07-09
📝 clean up, deduplication, and consistent formatting of CPC members

ctcpip:
normalizes the display of CPC members and also removes duplicate entries for voting members from the regular members list

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ CPC members should attend as many meetings as possible, and non-members are welc
 
 - Joe Sepi ([@joesepi](https://github.com/joesepi), IBM)
 - Matteo Collina ([@mcollina](https://github.com/mcollina), Platformatic)
-- Sara Chipps ([@sarajo](https://github.com/sarajo), LinkedIn)
+- Tobie Langel ([@tobie](https://github.com/tobie), UnlockOpen)
 
 ### Voting Members
 
--- a/README.md
+++ b/README.md
@@ -108,11 +108,11 @@ CPC members should attend as many meetings as possible, and non-members are welc
 
 ### CPC Chair
 
-Joe Sepi ([@joesepi](https://github.com/joesepi), IBM)
+- Joe Sepi ([@joesepi](https://github.com/joesepi), IBM)
 
 ### CPC Vice Chair
 
-Tobie Langel ([@tobie](https://github.com/tobie), UnlockOpen)
+- Tobie Langel ([@tobie](https://github.com/tobie), UnlockOpen)
 
 ### CPC Directors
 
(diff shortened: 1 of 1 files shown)


erickzhao (approved):

rxmarbles (approved):

README.md
```diff
@@ -170,11 +181,9 @@ Anyone who has been active in the foundation or one of its member projects, as d
 - Even Stensberg ([@evenstensberg](https://github.com/evenstensberg))
 - Jean Burellier ([@sheplu](https://github.com/sheplu))
 - Jon Church ([@jonchurch](https://github.com/jonchurch))
-- Jordan Harband ([@ljharb](https://github.com/ljharb))
 - Jory Burson ([@jorydotcom](https://github.com/jorydotcom))
 - Marco Ippolito ([@marco-ippolito](https://github.com/marco-ippolito))
 - Mert Can Altin ([@mertcanaltin](https://github.com/mertcanaltin))
-- Michael Dawson ([@mhdawson](https://github.com/mhdawson))
```
ljharb, line 177: this change isn't correct; the two memberships are distinct, and remain even if the voting members lose the election to be one.

README.md
```diff
@@ -185,12 +194,10 @@ Anyone who has been active in the foundation or one of its member projects, as d
 - Rich Trott ([@Trott](https://github.com/Trott))
 - Rifa Achrinza ([@achrinza](https://github.com/achrinza))
 - Ruy Adorno ([@ruyadorno](https://github.com/ruyadorno))
-- Sara Chipps ([@sarajo](https://github.com/sarajo))
```
tobie, line 188: Also Sara should stay in that list

README.md
ctcpip: oh?  some people who are voting members were not listed as regular members, so that's the thing that needs correcting then it sounds like

ljharb: Technically I think an impact project can appoint a voting member who's not a regular member, so that might not be wrong either.

README.md
ctcpip: whoops, I thought they were a voting member.  good catch.  (although it sounds like this is superseded by Jordan's comment anyway)

README.md
ctcpip: but there's language like this in governance:

> Note: Former Voting members whose terms have just ended will automatically become Regular members, unless they indicate otherwise.

ctcpip: so it sounds like technically you do not dual-card

ljharb: ah ok then yes, every voting member should also be in the regular member list :-)

ctcpip: while it doesn't seem to make any substantive difference, I guess it's easier to leave them in than to take them out and then have to re-add them later when folks may no longer be voting members so...  you got it, dude

ctcpip: @ljharb no rush, but this PR is still blocked on your change request

ljharb (approved):

PR #1336, opened 2024-07-14
Simplify proposal process

tobie:
The CPC is no longer using the formal proposal process outlined in the proposals directory so it made sense to retire it. Additionally, this PR:

- removes the artifacts from previous proposal process
- fixes all links to previous proposal to use permalinks instead


meetings/2019/2019-01-07.md
```diff
@@ -30,7 +30,7 @@ Extracted from **bootstrap-agenda** labelled issues and pull requests from the *
 * CPC Charter
   - 5 minute timebox
   - Refs:
-    - [Proposal Readme](https://github.com/openjs-foundation/cross-project-council/blob/HEAD/proposals/approved/CPC_CHARTER/README.md)
+    - [Proposal Readme](https://github.com/nodejs/bootstrap/blob/03ba5897e9c4e01ac2bf6b118a9a5ea2b5b85f61/proposals/mdawson-cpc-charter/README.md)
```
ljharb, line 33: shouldn’t this be https://github.com/openjs-foundation/cross-project-council/blob/a376c1dabb01fb1f0ac66affa9104876dee76fa9/proposals/approved/CPC_CHARTER/README.md ?

If the concern is archival it seems like we’d want to avoid relying on repos outside of this org

timmywil: Agreed. Seems strange that it links out.

tobie: My main concern is being able to cleanup the repo without making it really difficult to follow relevant links. I felt like the most reasonable decision was to pick the permalinks that people would have chosen back then. I do get your point however, although I feel like it's partially mitigated by Node being a foundation project.

tobie: This is a long-winded way of saying: you're right but I hate having to do that work. :D

ljharb: go to the original link, and hit the key shortcut for copying the permalink? (it’s in the drop down menu button on the upper right). You can also hit y and copy the resulting URL.

ljharb (approved):

timmywil (approved):

tobie:
Alright. I think I fixed all link issues now. Thanks for raising this, folks.

mhdawson:
I wonder if the past proposal have any context that should still be available somewhere?  Don't feel too strongly one way or the other but thought I'd mentioned it.

mcollina (approved):
lgtm

ovflowd (approved):
LGTM!

VoltrexKeyva (approved):

PR #1530, opened 2025-08-08
moved cosmos to at large

bensternthal:
All requirements have been met, this PR moves Cosmos from Incubating to At-Large.


tobie:
For ref., see https://github.com/openjs-foundation/project-status/issues/164.

PR #1654, opened 2025-09-25
Rename "Complete" project stage to "Feature-Complete", add new "Sunsetting" stage

tobie:
## Goals:

1. Provide more transparency and clarity around project lifecycle by distinguishing feature complete projects from sunsetting ones.
2. Make it easier for project to transition to "Feature-Complete" without necessarily implying that they're sunsetting.

### Q: Why rename "Complete" to "Feature-Complete"?
The name "Complete" wasn't super clear to folks and people weren't super thrilled of adopting it. I believe this new name makes it easier to understand what it means.

### Q: What does the new "sunsetting" stage changes?

It splits the "Complete" stage into two:
1. "Feature-Complete" which means that while you're not adding new features you have no intentions of winding the project down for now.
2. "Sunsetting" which means that while you're still maintaining the project you're planning to wind it down by a given a date.

### Q: What's makes you move from "Feature-Complete" to "Sunsetting"?
The only difference between the two is that you've announced an EOL date. That's it.

### DIFF
--- a/PROJECT_PROGRESSION.md
+++ b/PROJECT_PROGRESSION.md
@@ -27,7 +27,7 @@ flowchart TD
     F --> Graduated 
     subgraph Graduated
        direction LR
-       a([Stage: Impact]) <--> b([Stage: At Large]) <--> G([Stage: Feature-complete])
+       a([Stage: Impact]) <--> b([Stage: At Large]) <--> G([Stage: Feature-Complete])
     end
     Graduated --> J{EOL announced?}
     J --> |Yes| H([Stage: Sunsetting])
@@ -39,7 +39,7 @@ flowchart TD
 | Stage | New Feature Development | Vulnerability & Bug Fixes | EOL Announced | EOL Reached |
 |-------|:-----------------------:|:-------------------------:|:-------------:|:-----------:|
 | Active (At Large/Impact) | ✅ | ✅ | ❌ | ❌ |
-| Feature-complete | ❌ | ✅ | ❌ | ❌ |
+| Feature-Complete | ❌ | ✅ | ❌ | ❌ |
 | Sunsetting | ❌ | ✅ | ✅ | ❌ |
 | Archived | ❌ | ❌ | ✅ | ✅ |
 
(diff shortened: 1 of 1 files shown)


bensternthal:
@tobie overall this looks solid to me. And I believe we should audit our projects once this is merged, there are likely projects that should have one of the new statuses applied. (e.g. https://github.com/globalizejs/globalize/issues/949)

The only question I had was related to the flowchart, I was unsure what was meant by the following:

Screenshot 2025-09-26 at 8 12 45 AM

tobie:
Oh, thanks for catching this, folks! I thought I had a fixed it. That's a search and replace gone wrong! (`s/Complete/Feature-complete/`).

tobie:
@bensternthal [wrote](#issuecomment-3339124678):
> And I believe we should audit our projects once this is merged, there are likely projects that should have one of the new statuses applied.

Could not agree more. We really want to normalize being more transparent about where a project is in its lifecycle.

UlisesGascon (approved):

tobie:
**Notes from today's CPC call:**

This was presented as a way to make it easier for maintainers to be transparent about their intentions wrt to the maturity of their project.

In particular, this changes allows projects that are feature complete, but don't have an intention to sunset to say so, without suggesting that they are on a track for archiving their projects.

wesleytodd (approved):
Seems pretty good to me. Just wanted to drop [this link](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md) here because I think we should have tooling and metadata around this type of thing someday. I also hope that we can see @ljharb's work specifying this stuff get added to this when it is ready.

tobie:
Thanks @wesleytodd. Appreciate the thoughtful review and the reference to [PACKAGE-SUPPORT.md](https://github.com/nodejs/package-maintenance/blob/main/docs/PACKAGE-SUPPORT.md) which I'm adding to the [inventory of resources for the CRA](https://github.com/orcwg/cra-hub/blob/main/inventory.md) (tracking here: https://github.com/orcwg/cra-hub/issues/293).

PR #1675, opened 2025-10-27
Update CPC meeting calendar links

bensternthal:
Updated calendar links for CPC meetings in README.


tobie:
I think this makes sense to fast-track?

If you agree, please add a 👍 to this comment.

bjohansebas (approved):

Issue #1114, opened 2023-06-30
Make it clear that that a project joining the Foundation includes an IP transfer 


mcollina:
I had a call with another maintainer, who mentioned that our language could be more precise on what intellectual property is assigned to the Foundation.

Note that this is mentioned in
https://github.com/openjs-foundation/cross-project-council/blob/main/PROJECT_PROGRESSION.md#onboarding-checklist.

It's not spelled clearly what assets would be transferred.

tobie:
Same comment as in #1115. The only IP transferred are trademarks. We should spell this out.

ljharb:
In nvm's case, the `nvm.sh` domain name was also IP that was transferred, but to be clear that wasn't a requirement unless I wanted the Foundation to pay for the domain.

tobie:
The domain name isn't _technically_ "IP", it's really just an asset. Similarly, a book or a picture themselves aren't IP (much to the chagrin of folks trading NFTs or [spending $3M on a Dune manuscript](https://www.esquire.com/entertainment/books/a38815538/dune-crypto-nft-sale-mistake-explained/)). Transferring the ownership of such assets doesn't imply anything about the transfer of the underlying IP. The author of the book remains the copyright owner. The photographer continues to own the rights to the picture.

It is worth noting that if the domain name infringes on the trademark (as it does in your example), you're kind of forfeiting it too by transferring the trademark to the Foundation. Indeed, regardless of who remains the domain registrant, the right to use the trademarked term in a domain name belongs to the trademark owner, not the registrant.

bensternthal:
This is covered in the contribution agreement.

tobie:
Could we point to a publicly available copy of it?

bensternthal:
@tobie we already do via the onboarding checklist. However I made a small pr to make this clearer and link to a non google drive version: https://github.com/openjs-foundation/cross-project-council/pull/1682

PR #1682, opened 2025-10-30
Linked to updated contribution agreement stored in Github

bensternthal:
Improves on #1114 

Previously we linked to an outdated doc. I uploaded an example that has recent changes from legal. I also moved this item to be first in the checklist as it really is the first thing that needs to happen.


nektro (approved):

tobie:
You rock!

VoltrexKeyva (approved):

UlisesGascon (approved):

PR #1688, opened 2025-11-05
Add requirement to communicate appeals to all parties

mcollina:
Clarified that appeals must be communicated to all parties involved.

### DIFF
--- a/conduct/COC_POLICY.md
+++ b/conduct/COC_POLICY.md
@@ -127,6 +127,8 @@ If the appeal follows an incident report managed by a project, the project is ex
 
 Appeals do not delay or block the execution of decisions communicated in the response.
 
+Appeals must be communicated to all parties involved.
+
 [CoC]: https://code-of-conduct.openjsf.org/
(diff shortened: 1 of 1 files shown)


mcollina:
The accused is already aware that a complaint has been filed against them, and they might be surprised that a story they considered closed actually proceeded for months.

I'll produce an alternative PR that limits the time that the appeal team can use to produce a decision.

mcollina:
This is the alternative https://github.com/openjs-foundation/cross-project-council/pull/1690 that limits to 30 days the time that a decision must be communicated to all parties involved. This would give a total of 60 days after the 1st level decision has been made.

Trott:
> The accused is already aware that a complaint has been filed against them

They are? This would seem to be at odds with the requirement that the harmed party's identity must be kept confidential. There are going to be cases where revealing that an accusation has been made is going to absolutely reveal the identity of the harmed party.

I haven't been engaged in the conversations around this stuff, so I could be way off base, but I have to imagine that If at all possible, we should list out what "all parties" means in the two places it is used. I think there might be multiple interpretations going on.....

ctcpip:
> The accused is already aware that a complaint has been filed against them

they should not have been

PR #1690, opened 2025-11-05
Add timeline for communicating appeal decisions

mcollina:
Clarified the communication timeline for appeal decisions.

Alernative to #1688.

### DIFF
--- a/conduct/COC_POLICY.md
+++ b/conduct/COC_POLICY.md
@@ -120,6 +120,7 @@ Projects are expected to participate in good faith in the escalation process and
 ## Appeals
 
 Harmed individuals and accused individuals have thirty (30) days after a response was provided to appeal to the CoC Team a decision made in response to an incident report. Decisions may only be appealed on the grounds that the documented process was not properly followed or that it did not meet the requirements of this policy. If the CoC Team finds that the appeal has merit, it may decide to override the decision that was previously made.
+The decision of the CoC team must be communicated to all parties involved within thirty (30) days.
 
 If the initial report was already managed by the CoC Team, a different set of CoC Team members should be chosen to manage the appeal process. The CoC Team may choose to involve external people if needed.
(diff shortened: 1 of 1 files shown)


conduct/COC_POLICY.md
```diff
@@ -120,6 +120,7 @@ Projects are expected to participate in good faith in the escalation process and
 ## Appeals
 
 Harmed individuals and accused individuals have thirty (30) days after a response was provided to appeal to the CoC Team a decision made in response to an incident report. Decisions may only be appealed on the grounds that the documented process was not properly followed or that it did not meet the requirements of this policy. If the CoC Team finds that the appeal has merit, it may decide to override the decision that was previously made.
+The decision of the CoC team must be communicated to all parties involved within thirty (30) days.
```
Trott: Is there some way we can make this and the other use of "all parties" more specific? Who exactly will be informed? And if we can't do that, might it be better to provide a little more discretion by using "relevant parties" or something like that?

ctcpip (changes requested):
Given my experience in how long it sometimes takes to resolve these things, this is not going to be doable in all cases, for a number of reasons.  It's fine to specify an ideal (e.g. SHOULD instead of MUST), but whatever we do needs to be realistic.

tobie:
I kind of feel this is already covered by the [section on responsiveness](https://github.com/openjs-foundation/cross-project-council/blob/main/conduct/COC_POLICY.md#responsiveness):

> Every incident report, escalation request, or appeal will be promptly acknowledged upon receipt and a response provided in a timely manner. All parties involved should be kept informed and updated in a timely and efficient manner.

I also don't think committing to a fixed timeline is reasonable. OTOH, I do feel like we need to do a better job of keeping all involved parties up to date when there are appeals, including the project's governing body, the CPC, and the board.

PR #1699, opened 2025-11-12
✨ flatten CPC chair group and allow for up to three co-chairs

ctcpip:
The CPC charter and governance do not distinguish any roles, responsibilities, or privileges between the CPC chair and the CPC vice chair, so they are effectively co-chairs in everything but name.

This proposal flattens the group to simply be chairs, or co-chairs, who would be elected as a group.

More significantly, it raises the number of (possible) chairs to a maximum of three.

> [!NOTE]
> Full disclosure: I personally ran for the chair and co-chair elections last year and lost both.  I have not decided whether I will run again in the future.

> [!IMPORTANT]
> I am leaving this topic off the CPC meeting agenda for now because I will not be able to attend the next two weeks' meetings, and would like to be present if/when this topic is discussed.

### DIFF
--- a/CPC-CHARTER.md
+++ b/CPC-CHARTER.md
@@ -75,9 +75,9 @@ In this case the member will serve out the remainder of the term and must be re-
 The public portion of the CPC discussions and meetings will be open to all observers and members.
 
 The CPC shall meet regularly using tools that enable participation by the community.
-The meeting shall be directed by the CPC Chair or the CPC Vice Chair.
-Responsibility for directing individual meetings may be delegated by the CPC Chair or by the CPC Vice Chair to any other CPC member.
-The CPC Chair and CPC Vice Chair roles confer Voting member status, and will be selected by the CPC members through consensus or if necessary a vote as described in the section titled "Voting".
+The meeting shall be directed by the CPC Chairs.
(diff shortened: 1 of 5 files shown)


tobie:
So... a few thoughts:

1. I agree with the lack of practical distinction between the two roles. However, changing that requires also changing the [bylaws](https://bylaws.openjsf.org/), which adds significant overhead and is a distraction from more pressing issues that the board needs to tackle.
2. I do agree that having to decide whether you're running for chair, or vice-chair, or both is a little silly. My hunch is that we should fix this through how we vote, not by rechartering the whole org. For example, we could just sort all candidates and the one with the most voices is chair, the second one is vice-chair.
3. I think a third co-chair would be interesting to try (I've never seen it be very effective in other orgs, but never experienced it myself, so I'm a little curious about it).
4. We do want to watch out about over representation of chairs from one company or one project, though, so I think that's something to keep in mind.
5. I really appreciate the voice and support you bring to the CPC and would love to figure out ways to get people more involved and better recognized for their involvement. Maybe recognizing the role of collab space chairs more publicly would be a good first step.

ctcpip:
> This would require also changing the bylaws

there's a typo in the bylaws I would like to fix, so maybe a good opportunity  😅

PR #1773, opened 2026-06-02
Create SECURITY.md for security policy and reporting

lawrenceboivin79-spec:
Added a security policy document outlining supported versions and vulnerability reporting.


ljharb:
This repo isn't a software project and doesn't need a security policy, and it seems like the version numbers in the PR are thus hallucinated?

